PR #4, opened 2015-06-26
New documentation

gabrielhase:
Completely overhaul the documentation to provide a "from nothing to pro" way into Livingdocs.


gabrielhase:
@marcbachmann @peyerluk @caillou @rawdiggie I completely overhauled the documentation project since the one we had before was pretty chaotic and IMO wouldn't help a new developer that much.
This still has a lot of holes, but the very basics are layed out. You can provide review comments here if you like, but I'll merge it in any case (it's far better than what was there before). In any case you should familiarize yourself with this documentation because: WHENEVER YOU CHANGE SOMETHING FROM NOW ON AND IT WAS DOCUMENTED HERE, YOU ARE REQUIRED TO UPDATE THE DOCUMENTATION. So it makes sense to at least skim it so you know when you have to update something.

The plan for the future: I'll continue working on the documentation if time allows. This means moving all the content from the various wikis here. This will be the **only** documentation project for Livingdocs and documentation will be fully open. This also means that you shouldn't work on the wiki pages anymore. If you have such a case, please discuss with me first.
The only other place where some "documentation"-like content will be is our blog. The wikis of the various projects will in future all point to this project. The Readmes of the projects should be restricted to setup descriptions and documentation should link here.

Issue #6, opened 2015-07-10
The Server API documentation is missing


groie:
The frontpage of Livingdocs documentation (https://github.com/upfrontIO/livingdocs) has a link to the Server API. This link gives me 404.

peyerluk:
Thanks for the comment. The server is a closed source project and we have to move the documentation out of there.

groie:
Thanks!  :+1: Looking forward!

marcbachmann:
Will close this because we've moved the documentation into this repo. Here's the link: https://github.com/upfrontIO/livingdocs/blob/master/server/home.md

PR #7, opened 2015-07-10
Add Server API documentation

peyerluk:


peyerluk:
Fixes #6

peyerluk:
@gabrielhase @marcbachmann please review if everything is up to date.

server/editing_api_authentication.md
```diff
@@ -0,0 +1,51 @@
+# API Authentication
+### `POST /authenticate` Logging In
+
+#### Parameters
+
+| Property      | Type    | Default  | Description
+| ------------- | ------- | -------- | -------------
+| `username`    | string  | -        | **required**, utf8 encoded username, normally your email address
+| `password`    | string  | -        | **required**, utf8 encoded password
+| `expires`     | integer | 7200     | Access token expiration time in milliseconds
```
marcbachmann: We've removed that expires param

server/deployment.md
```diff
@@ -0,0 +1,69 @@
...
+```bash
+heroku config:set aws__secret_key="your-secret"
+heroku config:set aws__access_key="your-key"
+heroku config:set aws__buckets__images="livingdocs-images"
+```
+
+If you want to copy settings from other heroku instances you can use this heroku plugin: https://github.com/ddollar/heroku-config
+
+### Example Configuration
+(Those are not real passwords. so don't try to use them.)
```
marcbachmann: I hope that's really true ;)

server/integration_bridge.md
```diff
@@ -0,0 +1,23 @@
...
+
+#### authorize(access_token, callback)
+```
+  access_token: Access token string
+  callback: must be called with one of the following options
+    * callback(err) when an error occurs
+    * callback(null, {user: userObject, credentials: Object)
+    * callback() When unauthenticated
+```
+#### find({id}, callback)
```
marcbachmann: I guess we can remove that file

marcbachmann:
Do you like to delete the old gh wiki?

gabrielhase:
> Do you like to delete the old gh wiki?

Lets wait a little with that. We should go through it and make sure first that we took everything over. Then we should delete it.

server/editing_api_authentication.md
peyerluk: Do you want to update this before I merge?

PR #9, opened 2015-07-14
add missing comma in example

hypebeast:
A comma was missing in the example.

### DIFF
--- a/design/create_designs.md
+++ b/design/create_designs.md
@@ -133,7 +133,7 @@ Following you see a basic paragraph definition. It consists of a configuration p
 ```html
 <script type="ld-conf">
 {
-  "name": "myParagraph"
+  "name": "myParagraph",
   "label": "My very first paragraph"
 }
 </script>


gabrielhase:
:+1: thanks!

hypebeast:
One more question regarding the example component: The example contains a HTML comment, but with this comment the example isn't working.

```
<script type="ld-conf">
{
  "name": "myParagraph",
  "label": "My very first paragraph"
}
</script>

<!-- your html comes here -->
<p doc-editable="text">
  Placeholder text
</p>
```

Is this a bug?

gabrielhase:
Thanks for pointing this out. The problem is that the comment is also seen as a node and the code currently only takes the first node. But yes, it is a bug in any case :smile: I just created a ticket in our framework to fix this.
Note: you can use comments in other places than before the root HTML tag.

hypebeast:
Ah, thanks for the explanation :)

But maybe its better to remove the comment in the example until the 'bug' is fixed. I think it's a little bit confusing for a programmer who follows the guide and it doesn't work as expected.

gabrielhase:
It's already fixed. @marcbachmann is just adding a test and will push the update in about half an hour.

hypebeast:
Cool :+1:

marcbachmann:
@hypebeast Regarding your comment problem, what are you using to compile the design?
I thought the comment node must be the problem, but I can't reproduce it :sweat:.

marcbachmann:
@hypebeast can you try to compile the design like that:
```
npm install -g upfrontIO/livingdocs-manager#template-filter-comment-nodes
ldm design:build ./source ./dist # ./source should be the directory that contains the livingdocs design configuration file (config.json)
```

hypebeast:
Hi Marc,

I'm using `grunt dev` to compile the design. During the compilation everything is fine and no error occurs. But when I try to preview the design i get an empty screen.

This is the output when i compile the design is:

```
$ ldm design:build ./src ./dist
info build Design compiled...
```

When I now start the server with `grunt express watch` everything is fine and the preview is working fine.

When I compile the design with `grunt dev` I get an empty screen.

marcbachmann:
I've updated the build process to remove text & comment nodes. That might have been the problem with your template. I've published the new version of livingdocs-manager. So you should be able to install `grunt-livingdocs-design` again to make it work.

PR #19, opened 2016-02-07
Overhaul

gabrielhase:
* change navigation
* add ToC everywhere
* shorten main text
* add description for publish plugin


gabrielhase:
@peyerluk this greatly improves our documentation and trims old and untrue stuff (you forgot to remove the webhook documentation after you broke it by the way...).
Please take a look at it and give me feedback. Also take a look at the PublishPlugin documentation. I make some assumptions there on the public API of the `documentVersion` that you should review.

PR #28, opened 2016-04-28
Fix doc regarding the design upload with channels

mthpvg:
You cannot add a new design and set it as default **IF** the name of the design is different from the current design name.
You can add a new design and set it as default **IF** the name of the design is the same as the current design name (the only difference would be the version number)
You need to create a new user (new project, new channel) and then you can select a newly uploaded design.

For more informations:
https://github.com/upfrontIO/livingdocs-manager/issues/31
https://github.com/upfrontIO/livingdocs-design-timeline/pull/15

@marcbachmann  can you confirm ?
@gabrielhase @masone  FYI

### DIFF
--- a/design/upload.md
+++ b/design/upload.md
@@ -29,18 +29,28 @@ The tool will ask you to provide a host (the default is fine) and enter your ema
 
 Go ahead and fetch the URL by simply pasting it into a browser. You should see the JSON design configuration file.
 
-### Using the design in your account
+### Using the design
 
-In your Livingdocs Beta account you will not yet see your design. You first need to update the configuration.
+In your Livingdocs Beta account you will not yet see your design.
+
+#### Your first design
+
+As your design has probably a different name than the one attached to your current user, you need to create a new user:
+
+```bash
+grunt setup-admin-user
+```
+
+You will be asked to supply your username and password before the configuration can be updated. Then you will choose from a list your newly uploaded design. This will also set this design as the default design that will be used when creating new documents with this new user.
+
+#### New design has the same name as the current design only version number changed
 
 Add the design to your configuration with 2 terminal commands:
 
 ```bash
-ldm project:design:add --name bootstrap --version 1.0.0
-ldm project:design:default --name bootstrap --version 1.0.0
+ldm channel:design-version:add --name bootstrap --version 1.0.0 --channel 1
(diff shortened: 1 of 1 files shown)


design/upload.md
```diff
@@ -29,18 +29,28 @@ The tool will ask you to provide a host (the default is fine) and enter your ema
...
-### Using the design in your account
+### Using the design
 
-In your Livingdocs Beta account you will not yet see your design. You first need to update the configuration.
+In your Livingdocs Beta account you will not yet see your design.
 
-Add the design to your configuration with 2 terminal commands:
+#### Your first design
+
+As your design has probably a different name than the one attached to your current user, you need to create a new user:
```
marcbachmann: What, no?

marcbachmann: Or a new project.

mthpvg: From Ralph:

From step 1 to 3 we use the design named spaceline (version is 0.3.1). But in step 4 the name remains timeline.
@marcbachmann's assumption was correct. If a channel has been created, it is not intended to change the design name. The name parameter gets ignored by ldm.

Can we remove the channel 1 or set channel 3 as default ?
ldm doesn't support removing a channel or set a default channel
deleting a channel can be done with a HTTP call to the server -> https://github.com/upfrontIO/livingdocs-server/blob/master/app/features/projects/routes/channel_routes.coffee#L13-L19
setting the current design version (formerly know as default design) can be set via ldm -> ldm channel:design-version:current

marcbachmann: But that doesn't mean we have to add a new user? We just need to create a channel and add the design to it.
And end users have no way to access our server tasks. So that doesn't help anybody.

mthpvg: Ok, I got it.

At that time, I did not now how to change channel for the current user.

mthpvg:
Newly created channel has no workable metadata configuration: https://github.com/upfrontIO/livingdocs-server/issues/976

masone:
Thanks for the documentation effort :) - still requires confirmation from Marc?

mthpvg:
> Newly created channel has no workable metadata configuration: https://github.com/upfrontIO/livingdocs-server/issues/976

In this issue Lukas told that you need to define configuration manually and it can't be done with grunt tasks at the moment. It seems a bit complex to document.

PR #31, opened 2016-07-11
rename hosted->dokku

gabrielhase:


gabrielhase:
@masone please review

marcbachmann:
@gabrielhase thanks for updating.
@masone we've changed that because I've used `hosted.livingdocs.io` for rancher. I thought that's not used. Sorry for that.

masone:
Yes we figured ;) This is a valid workaround. Dokku 💣

PR #39, opened 2016-08-25
Implementation of new toc

mthpvg:
**WIP**

Related epic: https://github.com/upfrontIO/livingdocs-planning/issues/4


masone:
- I would `mv contribution-guidelines/guides/* contribution-guidelines` and then move the two documentation related guides to a `contribution-guidelines/documentation` folder
- move `contribution-guidelines/guides/WERE_HIRING.md` to root level
- rm `contribution-guidelines/guides/README.md`
- remove the livingdocs from `livingdocs-concepts` or we would need to be consistent and also do `livingdocs-contribution-guidelines`, etc.
- rename `public-documentation` to `public` (same for `core-documentation`) as I think it's pretty obvious you're in a documentation repository. as you suggested we could rename the repo to `livingdocs-documentation` (github sets up a redirect, so no links will break)

masone:
Otherwise good to move on from my side

mthpvg:
Queue:

- [ ] Extend the CSS guide: dos and don’ts
- [ ] Rename the livingdocs repository to livingdocs-documentation
- [ ] Contribution guidelines are not structured as the rest of the document

PR #40, opened 2016-09-02
Print API description

gabrielhase:


gabrielhase:
@nephridium please review.
This description combines API and data structure examples with explanations. It would be nice to have your apiary links in here too. Since this is the official Livingdocs documentation and everything here is public, we would though need to make the apiary documentation public as well. Is this possible?
I left a couple of "TODO"s in the text. Can you go over them and explain where you know more than me?
@peyerluk fyi

public/APIs/print_api.md
```diff
@@ -0,0 +1,517 @@
...
+        </article>
+    </getPreview>
+</articleUpload>
+```
+
+The authentication block is equivalent to the other requests. The `getPreview` block first defines the layout or template id for which a preview is requested (NOTE: you previously sent this to Livingdocs in the template or layout response, see above) and the type of preview that is requested. The type can be:
+- `articleJpeg`, gets the preview in text and image form (both)
```
nephridium: Currently, `articleJpeg` is the only one working on huGO (and the only one we're using on livingdocs). The other values are apparently not evaluated correctly by NewsNT.

(`articleText` worked previously and would return the same response, only without the image link, - and slightly faster since NewsNT wouldn't need to generate the JPEG).

public/APIs/print_api.md
```diff
@@ -0,0 +1,517 @@
...
+    </authentication>
+    <getLayouts>
+        <attribute name="publication" value="yourPublication"/>
+        <attribute name="publicationDate" value="02.09.2016"/>
+    </getLayouts>
+</articleUpload>
+```
+
+As you can see there are basically 2 parts: an authentication and the layout request.
+The authentication sends you a username and a password that you can use to authenticate Livingdocs with your middleware. You can currently only configure one user for all requests.
```
gabrielhase: But that only concerns the implementor hugo. We send this information and a decent implementor should check it.

public/APIs/print_api.md
```diff
@@ -0,0 +1,517 @@
...
+        <attribute name="password" value="yourPassword"/>
+    </authentication>
+    <getTemplates>
+        <attribute name="publication" value="yourPublication"/>
+        <attribute name="allowedOnCms" value="true"/>
+    </getTemplates>
+</articleUpload>
+```
+
+Again, we are using an authentication block. The second block defines the publication name for which you want to get the templates. Note that you don't get a specific publication date since templates are independent of any specific edition. The `allowedOnCms` parameter will always be `true` for calls from Livingdocs.
```
gabrielhase: `allowedOnCMS` is a parameter of the response, not the request. And my assumption is that we only get templates that are allowed for Livingdocs 😄  Is this correct?

public/APIs/print_api.md
gabrielhase: Again, in this documentation we do not care about the specific implementor and which of the different calls they actually implemented.

public/APIs/print_api.md
nephridium: We are requesting a list of templates that are 'allowedOnCms', so it's a request param. Originally it would send all available templates from NewsNT, which was extremely slow, so we added a filter that corresponds to a flag on NewsNT.

PR #43, opened 2016-09-12
docs(npm): Update release docs

mthpvg:
@marcbachmann This is a PR of you from the upfrontIO/guides which is now merged in upfrontIO/livingdocs


marcbachmann:
I guess we can merge that. It's better than what we currently have in there. But it's not yet complete.

PR #47, opened 2016-09-22
Update metadata description with text service

gabrielhase:
https://github.com/upfrontIO/livingdocs-editor/pull/1226


masone:
Thoughts: There is not much value in opening a planning issue when the planning process is circumvented. The issue was opened during the sprint, work started immediately. There is no planning, no estimation, etc.

If you can provide a quick pull request that you can finish within the sprint without jeopardising your sprint goal, I find it ok to not open a planning issue. No artificial overhead required :)

PR #55, opened 2016-11-17
docs(styleguide): custom directives

mthpvg:
@rawdiggie: do you remember that we said we wanted to do a techtalk about that :smile: ?
In the meantime let's make a documentation page about it, feel free to contribute !

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ Experimental:
 
 #### Livingdocs editor
 
+★ [Styleguide](./core/guides/livingdocs-editor/styleguide.md) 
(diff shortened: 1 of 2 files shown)


rawdiggie:
@mthpvg We also need to get rid of the old styleguide. Oh man :( We have an issue for that, right?

mthpvg:
Yes we do: https://github.com/upfrontIO/livingdocs-planning/issues/74

rawdiggie:
@mthpvg 👍  Thx man.

PR #57, opened 2016-11-23
Documenting docker stuff, cleaning up

masone:
- Cleaned up the docs on docker
- Moved some bits from the stripped down READMEs to the docs
- Started removing duplicate documentation from the archive
- Ended up removing the whole archive, there was not much value in the remaining 3 files


masone:
@mthpvg there is no more archive and no more duplicated docs anymore - any objections?

masone:
Oh and: What's the reason for keeping an archive if all information is properly filed in the new structure 🤔

peyerluk:
> Ended up removing the whole archive, there was not much value in the remaining 3 files

`ARCHIVE/upfrontIO_livingdocs/ideas/document-model.md` is still a good read :) It is a candidate to be implemented in the coming months (at least partially). 

Same goes for the design migrations file.

I moved the content of both of those into Google Drive under `Development`.

mthpvg:
Thanks for that :+1:

PR #58, opened 2016-12-07
Latest update on how to create an user with grunt

rebelact:


peyerluk:
Should we add a link to this in the readme?

masone:
Let's try to keep it as loosely coupled as possible. Links will get outdated and will need to be updated (which will be forgotten). I'd rather invest in further improving the structure of the documentation so everyone can find it easily

masone:
After a more thorough look - I see several problems with this change. For example, some of the information is not correct:

- It's `user-create`, not `create-user`
- `user-create` can also create admins. The real difference is, that one works with every (even remote) server API, while the `user-create-admin` only can run on the target machine and directly connects to the DB

Also I think the specifics about how we deploy at livingdocs should not be referenced from the public documentation. I would go for a more generic wording

Closing this PR and including an adapted version here https://github.com/upfrontIO/livingdocs/pull/68

masone:
fyi @rebelact

PR #60, opened 2016-12-09
Update not existing link

rebelact:
Licecap's repo seems to be nonexistent and the new repo's last commit is from August 2015, company's link seems to be the most appropriate


marcbachmann:
There's also https://github.com/justinfrankel/licecap
but your link is even better 👍

PR #63, opened 2016-12-09
Update hotfix-howto to use script

peyerluk:


masone:
As discussed, I don't like the duplication of the script in favor of replacing one step. Maybe I can help analyzing why `jq` didn't work?

peyerluk:
With help from @masone and @marcbachmann I just updated the jq script.

masone:
Thanks, I like that solution much better!

PR #66, opened 2017-01-05
docs: simplify the README

masone:
@mthpvg I drastically stripped down the README. I hope this is okay?

Reasoning: I found the README utterly confusing since it combined undocumented things in the same list as documented things. It was also very hard to scan visually and was duplicating the directory structure manually.

I have moved the missing documentation to a planning issue https://github.com/upfrontIO/livingdocs-planning/issues/361, so we don't loose track of it. 

We should aim for a good folder structure and speaking filenames instead of manually curating a structure in the README. Without duplication, we can prevent links from getting outdated.


README.md
```diff
@@ -1,221 +1,9 @@
...
-
-  This part is aimed at developers who want to contribute to Livingdocs core repositories.
-
-
-#### Get to know the Livingdocs basics
-
-- [Introduction into Livingdocs](./concepts/introduction.md)
-
-
-## Public
```
marcbachmann: Can we move some of those descriptions into `/public/README.md`?

masone: I like the idea to have a README in subfolders as well. I don't think there is much value in the content we already have though. 

1. Manual duplication of the directory structure (no value). I'd rather be more concise on the naming than duplicating the information
2. Missing documentation (confusing to outsiders)

masone: Damn, I started myself a review with that comment 🙄

marcbachmann: If a doc directory has subfolders we should at least have some kind of description in the the readme of this directory. I don't think that this is a duplication. It's a summary about the responsibility/content.
I agree that we shouldn't go into that detail at all.

e.g. the description in here about the migration feature is probably enough https://github.com/upfrontIO/livingdocs/pull/66/files#diff-04c6e90faac2675aa89e2176d2eec7d8L78

masone: I like that! Will look into improving that in this PR, thanks for the feedback

masone:
Thanks for the feedback!

PR #73, opened 2017-01-25
Design restructuring

masone:


public/guides/design/README.md
```diff
@@ -29,23 +13,9 @@ Basic Information:
...
-"assets": {
-  "css": ["./stylesheets/boilerplate.css"]
-},
-```
-
-
 ## Components
 
-Inside of the folder `src/components` you can define HTML components that can be used in documents that use your design. These components will be automatically inserted into `dist/design.json` when you run `grunt build`.
+Inside of the folder `src/components` you can define HTML components that can be used in documents that use your design. These components will be automatically inserted into `dist/design.json` when you run a build.
```
marcbachmann: We have to mention https://www.npmjs.com/package/livingdocs-manager somewhere.
Maybe we can publish the webpack build plugin as separate module.

Do you want to mention that in the design process issue... or create a new one.

masone: There is a guide on uploading a design with ldm and the readme of ldm has additional information. I think it should be enough.

Feel free to open / extend issues as you see fit

PR #75, opened 2017-02-14
docs(api): Routing system

vhf:


core/livingdocs-server/rest-apis/public-api/routing-system.md
```diff
@@ -0,0 +1,130 @@
+# Routing API
+
+## Use case
+
+The **delivery app** is asked to deliver a **document** located at a **path**. To resolve a path to a document, the delivery app uses the Routing API endpoint.
+
+1. Delivery **app** receives `GET /article/adventure/i-m-on-the-road-again-09fed8` on the channel `web`
+1. Delivery **app** does `GET https://delivery-server/api/routing/web?path=/article/adventure/i-m-on-the-road-again-09fed8`
```
masone: Suggestion to be a tad more explicit (I skipped the host in the url when reading it): `Delivery **app** calls the public api on< http...`

core/livingdocs-server/rest-apis/public-api/routing-system.md
```diff
@@ -0,0 +1,130 @@
...
+1. Delivery **server** response looks like:
```
masone: Do you think `public api` would be clearer than `delivery server`?

core/livingdocs-server/rest-apis/public-api/routing-system.md
```diff
@@ -0,0 +1,130 @@
...
+    }]
+    ```
+
+1. (Down the road, the delivery **app** will ask the delivery **server** for an HTML rendering of the latest publication of this document and serve it to the client)
+
+## Architecture
+
+The routing API is composed of three parts:
+
+1. The *routing metadata plugin*,
```
masone: I get a bit confused by the use of bold and italic - it seems inconsistently used (above you use bold for the **delivery app**, below you use italic *delivery app*). 

It feels like it's not really adding to the readability too. Not sure if that's related to the inconsistency.

masone: Another example: `Routes indexer` is not italic, but the other points in the list are - I don't get why on first sight

core/livingdocs-server/rest-apis/public-api/routing-system.md
```diff
@@ -0,0 +1,130 @@
...
+Routing configuration takes place at two levels:
+
+1. Environment - globally configure and enable the routing, and
+1. Per document type per channel - enable routing for each document type for each channel you'd like to route to.
+
+#### Environment
+
+In your environment configuration file (here, `local.coffee`), a top-level config key named `routing` is required, with the following possible keys:
+
+```coffee
```
masone: The following is just a thought - I haven't a solution for this yet.

The more specific we are in the documentation, the more likely it will break if we do changes. I have seen many parts of documentation rendered unusable because it was not updated. Seeing very specific things like configs wakes that memory.

I'm still trying to find a rule for a middle way - how to write documentation that does not get outdated so quickly - be as specific as needed to be understandable, but be as generic as possible at the same time

masone:
Oh - and once merged, please send it to the team and ping Dibran and Michael - as they are working with it already :)

PR #76, opened 2017-02-15
Add configuration for livingdocs-server

peyerluk:
I start here with a documentation of the `livingdocs-server` configuration.

The main motivation is that many configurations have multiple options which are difficult to communicate in a code file. In addition to that the configurations are often spread across multiple environments and files and folders what makes it hard to know where to look for available configuration options. And there are many configurations which warrant additional explanations.

### DIFF
--- a/public/livingdocs-server/configuration.md
+++ b/public/livingdocs-server/configuration.md
@@ -1,0 +1,278 @@
+# Livingdocs Server Configuration
+
+## Services
+
+#### Server
+
+```coffee
+server:
+  port: 9090
+  max_json_request_size: '100kb'
+```
+
+
+#### Editor
+
+Configure the Editor Host. This is used for password reset & login urls.
+
+```coffee
+editor:
+  public_host: 'http://localhost:9000'
+```
+
+
+#### Postgres
+
+```coffee
+db:
+  host: 'localhost'
+  port: '5432'
+  user: 'postgres'
+  database: 'li_local'
+  pool_min: 1
+  pool_max: 5
+  ssl: false
+```
+
+#### Amazon S3
+
+```coffee
+aws:
+  force_local_storage: false # defaults to `false`
+  secret_key: 'yourSecretKey'
+  access_key: 'yourAccessKey'
+```
+
+#### Pusher
+
+Disabled by Default. Enable to show how is viewing documetn and allow for basic
+collaboration features in the editor.
+
+```coffee
+pusher:
+  enabled: false
+  app_id: '108378'
+  key: 'dda3c0ca58ed2f649ea8'
+  secret: '*****'
+```
+
+
+#### Logging
+
+Possible log levels: `['debug', ''success', 'warn', 'error']`
+
+```coffee
+loglevel: 'success'
+```
+
+
+## Features
+
+
+#### Authentication
+
+Configure the authentication feature.
+
+*Important* Replace `access_token_secret` and `connections.local.config.secret`
+with a randomly generated string for every environment you use.
+Changing the `access_token_secret` will render all access tokens in the
+database useless.
+
+##### `connections.local`
+
+Represents the default email/password authentication strategy.
+
+- `default_login_domain` Allows to omit the email suffix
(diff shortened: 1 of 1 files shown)


gabrielhase:
Personally, I think this approach is prone to fail. The configuration options are changing quite quickly and this file will become outdated or even worse, wrong. And wrong documentation is worse than no documentation.

What I tried to do, is to document the configuration in the boilerplate projects directly in code: https://github.com/upfrontIO/livingdocs-server-boilerplate/blob/master/conf/environments/all.js

I always use the `all.js` file there. Since it's a boilerplate there are no environments, e.g. the environment is just empty. So it's not spread around many files.

Also, whenever you update the boilerplate with a new version and a configuration breaks, you need to change it in the `all.js` file. So then you have a fighting chance that you will actually also change the documentation that is set right next to it. At least better than if it's in a separate file.

What do you think?

public/livingdocs-server/configuration.md
```diff
@@ -0,0 +1,278 @@
...
+    watch_interval: 1000 # (default: 1000)
+  db: 'leveldown' # (default: leveldown) to use leveldb, npm i leveldown and set 'leveldown' here
+  # db_options: # (optional) cf. https://github.com/Level/levelup#options
+
+```
+
+
+#### Search
+
+Configure the elasticsearc instance used by the search feature.
```
marcbachmann: Typo

peyerluk:
I think documentation of the server configuration is one of the most essential documentations we have. And I assume it will be simpler to keep it up to date than to maintain multiple configurations with all possible configuration options in the downstream (which would be required to reflect the different options we have). This makes it also much easier to see for example all options one has for e.g. the authentication feature. And creating a doc is easier than creating a running configuration which also would need to be tested.

In our setup changing and extending configuration options is always a big thing. Multiple people have to be involved into the decision and it must be communicated to even more people to improve  our return on investment. So I think having an explicit configuration is more a plus to that process than a hassle.

peyerluk:
Another thing to consider is the testing strategy. I would propose to test the different configuration options in `livingdocs-server` project itself. And not delegate that responsibility to the boilerplate project. We're already working into that direction by making the feature code in the server stateless so we can instantiate it with different configurations in unit and integration tests. But there are other options we could start talking about.

masone:
I generally agree with gabriel. That's something we *will* have to handle though. I'd rather have it in once central place than scattered around the documentation. For example @vhf started documenting routing specific config somewhere else. 

This way at least you know where to update it. Starting with documenting it is not a bad thing. It will enable us to learn from it. I won't  stand in the way of merging this, even though I share the above concerns.

masone:
What if we could generate the documentation? If we had some wrapper around the config object, we could add additional information to each parameter

peyerluk:
> What if we could generate the documentation?

That's a good idea. But it involves creating all the comments and infos _and_ creating a dsl _and_ creating a process to create a documentation from it. I'd like that but it's too much for our next step.

> For example @vhf started documenting routing specific config somewhere else.

He has a pull request that adds it here to this project, right? I think that would be an argument for having the config docs here as well. I like what you have done in the boilerplate but it's the first time I hear about it.

DaRaFF:
My first thought was also to automate the generation of the documentation somehow.
I have an idea in mind which I want to share.

### Add Documentation for a Config Value and Validate the Config
Currently a config value will be registered with `config.set('key', 'value')`. We could extend the registry with `config.setDocumentation('key', 'documentation...')`. Then we could create some scripts which validates if a config key has a documentation and spits out a warning if not. Another script could generate a documentation from a specific node.


### Move Configs to the Features
It's not a solution to the mentioned problem, nevertheless it should improve the current problem with this BIG configs. We should start to move the configurations to the features itself. Then it's pretty clear where the config belongs to, it's easier to assign, test, add, remove and document configs.

peyerluk:
This change is now part of https://github.com/upfrontIO/livingdocs/pull/78

So I'm closing this as duplicate.

PR #78, opened 2017-02-16
New information architecture

peyerluk:
#### Goals

- Only cater to one audience. Customers
- Use «Concepts» to explain the main topics. The goal in this section is to give a high level overview about the general motivations and use cases of a part of Livingdocs
- Use «Reference Documentation» to explain all possible options of a feature or Api.
- Group all Walkthroughs together.
- Get rid of the ambiguous term «guides» as we use it now. Replace it with walkthroughs which always need to contain clear step by step instructions with a clearly defined end goal.
- Add a glossary
- Rewrite examples from coffee to javascript


README.md
```diff
@@ -1,9 +1,51 @@
...
-Livingdocs is not only a CMS but a component-based solution to publish across many devices and products.
 
-This technical documentation is split up in the following main parts:
-* **Concepts:** The fundamental concepts and philosophy behind Livingdocs. This helps you understand how we at Livingdocs think about the software. You will also learn the right words to describe specific things
-* **Contribution guidelines:** Describes how we collaborate and contains details on best practices we follow
-* **Public:** In the public documentation we cover everything you need to know for using a custom Livingdocs instance. **Audience: service customers**
-* **Core:** Detailed inventory of APIs, configurations, tools and operations available for each core project: server, editor and framework. It is the internal/private documentation. It should not overlap with the public documentation, but together, they hold a complete documentation of the Livingdocs codebase. **Audience: integration customers, core-developers/designers**
+## Welcome
+
+- Getting Started
```
gabrielhase: All of those should be links

gabrielhase:
> Only cater to one audience. Customers

I don't fully agree with this. Just make 2 sections, 1 for customers (public), 1 for core devs (core). The documentation should cater to both and the differentiation should be fluid, i.e. customers that heavily use Livingdocs can start reading the core stuff.

peyerluk:
In my opinion the 2 sections do not work. If you look in detail at how the files are structured currently this becomes quickly apparent. There is not much in our docs that is targeted at core devs. And the separation does not make it clear what our apis are. Currently a lot of stuff that is for customers is in the core folder. The majority of files currently in core actually belongs into public. Following the rule that every breaking change should have a corresponding change in the public docs.

gabrielhase:
Ok, I guess then we have a similar understanding. I was just concerned that you want to delete the files under `core`

masone:
No worries! Files under core will *not* be deleted :)
Thanks for the push here lukas. This is really hard to review and prone to merge conflicts. I would go for an early merge - we can always work on the details later

masone:
Thanks for picking up the work again here! 🌮  🌮  🌮

PR #80, opened 2017-03-07
Explain component configuration

gabrielhase:


gabrielhase:
@masone @peyerluk I'll merge this directly since the previous state was a disaster. I don't know how this ended up in the documentation, but literally nothing that was written corresponds to any truth. Having such documents in the documentation is way worse than having no documentation at all. We definitely need better QA processes on documentation. Would be great if you could track down how this ended up here.

peyerluk:
Das war ein Konzept für die Zukunft. Das ist wohl aus versehen in der Doku gelandet beim umstrukturieren. Ich hatte so Zeugs mal in einem entsprechend benannten Ordner.

PR #83, opened 2017-03-10
Document more editable options

gabrielhase:


marcbachmann:
editable.js is open source and its documentation belongs in there if it isn't linked to this repo.

gabrielhase:
> editable.js is open source and its documentation belongs in there if it isn't linked to this repo.

Only the stand-alone configuration.
What I am documenting here is the configuration that you can set in the editor, i.e. the ones that customers can customize in their own editor boilerplates.

PR #87, opened 2017-03-30
Update docker.md

mthpvg:
@masone: Fixing tiny things.

### DIFF
--- a/core/guides/deployment/container/docker.md
+++ b/core/guides/deployment/container/docker.md
@@ -69,7 +69,7 @@ Create a configuration file for the environment `dokku_<name>`. Point your edito
 Then you are ready to deploy by simply pushing to your remote:
 
 ```
-git push dokku master
+git push <name> master
 ```
 
 Your editor is available on `http://<name>.hosted.livingdocs.io`


core/guides/deployment/container/docker.md
```diff
@@ -32,7 +32,7 @@ ssh -t dokku@dokku.livingdocs.io lvds:create <name>
 Add a git remote in your local server repository:
 
 ```
-git remote add <name> dokku@dokku.livingdocs.io:<name>-server
+git remote add <name>-server dokku@dokku.livingdocs.io:<name>-server
```
masone: this is not necessary. the remote can have any name, but yours is more concise. I'm afraid that the instructions now fail for people who already started just using `staging` instead of `staging-server` as remote

If you change this, we would have to instruct everyone to either update their remotes names or rethink when reading the docs

I'd vote for not doing this change

core/guides/deployment/container/docker.md
```diff
@@ -44,7 +44,7 @@ ssh -t dokku@dokku.livingdocs.io config:set <name>-server auth__access_token_sec
 Then you are ready to deploy by simply pushing to your remote:
 
 ```
-git push <name> localbranch:master
+git push <name>-server master
```
masone: you're removing an important bit of information. if you push from a local branch other than `master`, let's say `localbranch` with `git push origin master` it would push the local `master`, not your `localbranch`

`git push <name> localbranch:master`

core/guides/deployment/container/docker.md
```diff
@@ -69,7 +69,7 @@ Create a configuration file for the environment `dokku_<name>`. Point your edito
 Then you are ready to deploy by simply pushing to your remote:
 
 ```
-git push dokku master
+git push <name> master
```
masone: this one I agree with! we should change it to `git push <name> localbranch:master` though

mthpvg:
I reverted the two changes as asked and kept the relevant one.

PR #88, opened 2017-03-30
docs: proxy in architecture

masone:


public/guides/hosting/proxy.md
```diff
@@ -0,0 +1,10 @@
+## Running Livingdocs behind a proxy
+
+Where applicable, Livingdocs respects the environment variables `HTTPS_PROXY` and `HTTP_PROXY` by default and routes all external requests through the given proxy. 
+
+```
+HTTPS_PROXY=https://my.proxy.com
```
marcbachmann: we need to use `HTTPS_PROXY=http://my.proxy.com` to get it to work

marcbachmann: That probably depends on the certificate the proxy uses. If it uses https and has a valid ssl certificate, it should work.

masone: `HTTPS_PROXY` seems to have been working at bluewin at least

PR #98, opened 2017-05-17
List mail server as external service

masone:
- list mail server as external service
- stumbled upon some duplication regarding the external services that is now removed


setup-and-deployment/external-services.md
```diff
@@ -0,0 +1,57 @@
+## External services
+For specific features, livingdocs uses external services. All external systems expose a REST interface, so only ports 443 and 80 are required.
```
marcbachmann: The mail server isn't limited to that. That's configuration-specific.

masone: right!

PR #101, opened 2017-06-06
docs(api): Routing system

vhf:
Continuation of https://github.com/upfrontIO/livingdocs/pull/75

@masone I took your review into account.

@peyerluk Please check that this file is at the right location, a bunch of things moved around since that PR was first opened.

### DIFF
--- a/reference-docs/server-public-api/routing-system.md
+++ b/reference-docs/server-public-api/routing-system.md
@@ -1,0 +1,141 @@
+# Routing System
+
+## Use case
+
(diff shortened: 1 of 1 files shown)


reference-docs/server-public-api/routing-system.md
```diff
@@ -0,0 +1,141 @@
+# Routing API
```
peyerluk: I'd prefer to just title it `Routing` or `Routing System` as the filename. As I see the routing as a part of the public api.

PR #102, opened 2017-06-12
Logging in @livingdocs/server

peyerluk:


reference-docs/server-configuration/logging.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+A few examples how to use the logger:
+```js
+log.debug('I never expected this to happen.')
+log.info({foo: 'bar'})
+log.info({foo: 'bar'}, 'Logging an object.')
+log.info({foo: 'bar'}, 'And a message %s', 'with interpolation')
+log.warn('Requiring lib/something is deprecated.')
+log.error(new Error('noooo!'), 'ouch!')
```
vhf: We should explicitly mention that only the first of the 11 possible arguments of a logging function can be an object, and that this object should be sanitized beforehand: it should not have any of the following root-level keys: `pid`, `hostname`, `level`, `time`, `msg`, `v`

vhf: Perhaps adding a few "bad examples" would help, something like:

```js
log.info('foo', {foo}) // object is not in first position
log.info({time: new Date()}) // `time` is one of the 5 forbidden keys
```

reference-docs/server-configuration/logging.md
```diff
@@ -0,0 +1,85 @@
...
+--------|-------
+'FATAL' | 60
+'ERROR' | 50
+'WARN'  | 40
+'INFO'  | 30
+'DEBUG' | 20
+'TRACE  | 10
+
+
+#### Example Request Log:
```
vhf: Adding an example showing how to log a request might be helpful

reference-docs/server-configuration/config.md
```diff
@@ -61,15 +61,25 @@ pusher: {
...
 
-
 #### Logging
 
-Possible log levels: `['debug', ''success', 'warn', 'error']`
-
 ```js
-loglevel: 'success'
+logs: {
+  enabled: true,
```
marcbachmann: That doesn't make much sense. You always want to show errors. Why would you disable logging?

marcbachmann: Ah, maybe in the tests? 😆

reference-docs/server-configuration/logging.md
```diff
@@ -0,0 +1,85 @@
...
+
+## How to use the logger
+
+The logger is exposed on the `liServer`:
+```js
+const conf = require('./conf')
+const liServer = require('@livingdocs/server')(conf)
+
+liServer.log.info('logging')
+```
```
vhf: IMO this example gives the impression that developers should instantiate a `liServer` whenever they want to get a hold on logging methods.

Warning about this might be a good idea.

reference-docs/server-configuration/config.md
```diff
@@ -61,15 +61,25 @@ pusher: {
...
+
+  // possible log levels: 'error', 'warn', 'info', 'debug', 'trace'
```
vhf: Isn't `fatal` another possible log level? It's referenced in `logging.md`

vhf:
I think it would be good to have some logging guidelines. For instance: do we log stack traces? Or the file and line? Or the name of the module + function logging something?

reference-docs/server-configuration/logging.md
vhf: * We should also be explicit about the issue of logging `Error` subclasses that have any properties named  `pid`, `hostname`, `level`, `time`, `msg`, or `v`.
* And state that logging an `Error` only works if the `Error` is not in an object, e.g. `log.error({x: new Error()})` will log `"x": {}`

peyerluk:
@vhf Thanks for the comments. That was helpful to see my own biases!

And regarding the logging guidelines. This is something we will talk about going forward. Let us pay attention in code reviews if the logged information makes sense and if we can identify good patterns.

reference-docs/server-configuration/logging.md
```diff
@@ -0,0 +1,101 @@
...
+
+The values `pid`, `hostname`, `level`, `time` and `v` are always added by the logger.
+`msg` will be the message. Any additional keys are additional information that varies
+depending on what is logged.
+
+The level is a numeric value.
+These are the values of the log levels you will see in the logs:
+
+Level   | Value
```
DaRaFF: Do we really need so much log levels? I know it's a well known standard, but I always talk with people and they are usually puzzled what kind of log message to put on which log level (including me).

Maybe a small additional description of the different states would help?

reference-docs/server-configuration/logging.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Example json log:
+
+```js
+log.warn({foo: 'bar'}, 'A log message.')
+```
+
+```json
+{
+  "pid": 40073,
```
DaRaFF: What do you think about additional standard fields like

- Docker Container Name
- Feature Name
- File / Line Number

reference-docs/server-configuration/logging.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+The request logs have the same format as any other logs.
+However they have the additional keys `req`, `res` and `responseTime`.
+
+In cases when our api returns a standard error in the body these will be
+added to `res.error`.
+
+`req.id` represents the requestId. The requestId can be overwritten in the request by
+sending an `X-Request-Id` header (This works independently from the logging).
```
DaRaFF: You mentioned that we also should/could use a log aggregator for local development. Maybe it's worth to add a description how to use such an aggregrator?

reference-docs/server-configuration/logging.md
```diff
@@ -0,0 +1,101 @@
...
+  "msg": "request completed",
+  "req": {
+    "id": "72ec31ac-baab-4e29-bfc0-a91a57d485ee",
+    "method": "GET",
+    "url": "/users/me"
+  },
+  "res": {
+    "statusCode": 200
+  },
+  "responseTime": 41,
```
marcbachmann: Res.ContentLength is also important

reference-docs/server-configuration/logging.md
peyerluk: Yes, you have to instantiate a `liServer` first. But what that means should be documented elsewhere in my opinion. It's not part of the logging docs. Whenever you want to access an instance variable you have to think about how to get hold of the instance – creating a new one or getting hold of an existing instance.

reference-docs/server-configuration/logging.md
peyerluk: - Docker Container Name

I don't think the container name is accessible from inside the process.

- Feature Name, File / Line Number

These we can do in the future if we like.

reference-docs/server-configuration/logging.md
peyerluk: These are all the ones supported by pino. Not that much we can do about it as long as we use pino.

And yes, we can work on short guidelines for each level.

reference-docs/server-configuration/logging.md
peyerluk: I think its corret not to add id as it is not part of the response. Plus this way it is not overridden if we define a custom serializer (which we do).

reference-docs/server-configuration/logging.md
peyerluk: Good Point. Will do at a later point.

PR #114, opened 2017-08-03
WIP: Explorations for new features

peyerluk:


reference-docs/new-design-and-channel/channel.js
```diff
@@ -0,0 +1,81 @@
+const channelConfig = {
+  handle: 'web',
+  label: 'Web',
+  description: '...',
+  language: 'de',
+
+  enabledFeatures: [
```
marcbachmann: Exactly what I wanted. But every feature should should need a configuration.
Do you want to have them in the root and namespaced?

We could also use an array or object with feature name and config. 

```
enabledFeatures: [{
  name: 'documents',
  contentTypes: [...]
}, {
  name: 'menus',
  ....
}]
```

gabrielhase:
can we close this?

PR #120, opened 2017-08-24
docs(hooks): Document documents and render hooks

vhf:
Related to planning issue 1348

### DIFF
--- a/reference-docs/server-configuration/channel-config.md
+++ b/reference-docs/server-configuration/channel-config.md
@@ -138,7 +138,7 @@ copy: [
 ```
 
 
-## Hooks: before render, on publish, on unpublish
+## Hooks
 
 A few APIs are provided to allow hooking into the document (un)publication process and into the rendering pipeline.
 
@@ -160,7 +160,12 @@ Examples for each hook type:
 
 ### Publish/Unpublish Hooks
 
-These two are set at once on the `documents` feature.
+These two are set at once on the `documents` feature:
+
+* `publishHook`:
+    * `({documentType, {documentVersion, renditions}}, callback)`
+* `unpublishHook`:
+    * `({documentType, {documentVersion}}, callback)`
 
 ```js
 const server = require('livingdocs-server/core')
@@ -188,7 +193,7 @@ liServer.registerInitializedHook((done) => {
       documentType,
       payload // payload is {documentVersion}
     }, callback) => {callback()}
-  })
+  }, done)
 })
 
 liServer.listen(port, function (err) {
@@ -197,14 +202,14 @@ liServer.listen(port, function (err) {
 })
 ```
 
-* `publishHook`:
-    * `({documentType, {documentVersion, renditions}}, callback)`
-* `unpublishHook`:
-    * `({documentType, {documentVersion}}, callback)`
-
 ### Before Render Hooks
 
-This one hooks into the `render-pipeline` feature. Here is a full example including server initialization:
+This one hooks into the `render-pipeline` feature. The `beforeRenderHook` is the last opportunity you have to modify/transform a document before it bets rendered.
+
+* `beforeRenderHook`:
+    * `({documentType, rendition}, callback)`
+
+Here is a full example including server initialization:
 
 ```js
 const server = require('livingdocs-server/core')
@@ -217,7 +222,7 @@ liServer.registerInitializedHook((done) => {
   liServer.features.api('li-render-pipeline').registerRenderHooks({
     projectHandle: 'your-interesting-project',
     channelHandle: 'some-channel',
-    beforeRenderHook: ({documentType, rendition}, callback) => { // here `payload` is a rendition
+    beforeRenderHook: ({documentType, rendition}, callback) => {
       if (['interview', 'biography'].includes(documentType)) {
         liServer.log.info("We're about to render something about somebody!")
         // do something with the rendition:
@@ -229,9 +234,7 @@ liServer.registerInitializedHook((done) => {
 
       callback()
     }
-  })
-
-  done()
+  }, done)
 })
 
 liServer.listen(port, function (err) {
@@ -240,8 +243,50 @@ liServer.listen(port, function (err) {
 })
 ```
 
-* `beforeRenderHook`:
-    * `({documentType, rendition}, callback)`
+### Document Lists Hooks
+
+Two hooks are available in the `document-lists` feature.
+
+* `updateListHook`:
+
+        ({
+          listId,
+          remove: [30, 199, …],
+          add: [{id: 77, order: 12}, …]
+        }, callback)
+
+* `getListHook`:
+    * `({listId}, callback)`
+
+Here is a full example including server initialization:
+
+```js
+const server = require('livingdocs-server/core')
+const config = require('livingdocs-server/conf')
+const liServer = server(config)
+const port = liServer.config.get('server:port')
+
+
+liServer.registerInitializedHook((done) => {
+  liServer.features.api('li-document-lists').registerListHooks({
+    projectHandle: 'your-interesting-project',
+    channelHandle: 'some-channel',
+    updateListHook: ({listId, remove, add}, callback) => {
+      console.info(`${listId} update happening, removing ${remove.length} things, adding ${add.length} things.`)
+      callback()
+    },
+    getListHook: ({listId}, callback) => {
+      console.info(`Getting list ${listId}`)
+      callback()
+    }
+  }, done)
+})
+
+liServer.listen(port, function (err) {
+  if (err) throw err
+  liServer.log.info('Listening on http://localhost:%s, started within %ss', port, process.uptime())
+})
+```
 
 ## A visual example of the `editor.frontend` configuration
 
--- a/reference-docs/server-configuration/channel-config.md
+++ b/reference-docs/server-configuration/channel-config.md
@@ -142,59 +142,65 @@ copy: [
 
 A few APIs are provided to allow hooking into the document (un)publication process and into the rendering pipeline.
 
-Although these hooks should preferably be registered before the server gets initialized, it is also possible to register them at runtime using the same APIs. (This is particularly handy for testing purpose, but also useful if you create projects or channels at runtime and need to set hooks for these.)
+Although these hooks should preferably be registered before the server gets initialized (using [Server Initialized Hooks](../../reference-docs/server-extensions/server-initalization.md#initialized-hooks)), it is also possible to register them at runtime using the same APIs. (This is particularly handy for testing purpose, but also useful if you create projects or channels at runtime and need to set hooks for these.)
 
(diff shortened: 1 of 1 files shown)


reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,98 @@ copy: [
...
+
+Examples for each hook type:
+
+### Publish/Unpublish Hooks
+
+These two are set on the `documents` feature.
+
+```js
+// publish hook
+liServer.features.api('li-documents').hooks
```
peyerluk: can we drop the `hooks` and just do `liServer.features.api('li-documents'). registerPublishHook()`? Then it would be the same as in the beforeRender hook.

peyerluk: And what about grouping it a bit?

Like this:
```js
.registerPublishHooks({
  projectHandle: 'your-awesome-project',
  channelHandle: 'some-channel',
  publishHook: () => {},
  unpublishHook: () => {}
})
```

reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,98 @@ copy: [
...
+
+This one hooks into the `render-pipeline` feature. Here is a full example including server initialization:
+
+```js
+const server = require('livingdocs-server/core')
+const config = require('livingdocs-server/conf')
+const liServer = server(config)
+const port = liServer.config.get('server:port')
+
+liServer.features.api('li-render-pipeline').registerBeforeRenderHook({
```
peyerluk: Here we could do 
```js
.registerRenderHooks({
  projectHandle: 'your-interesting-project',
  channelHandle: 'some-channel',
  beforeRenderHook: () => {}
})
```

reference-docs/server-configuration/channel-config.md
vhf: What about renaming `registerPublishHooks` to `registerPublicationHooks` in this case?

I'd find it less confusing. It would be the `unpublishHook` of the publication hooks instead of the `unpublish` hook of the `publish` hooks.

vhf: What behaviour would expect if you did 

```js
.registerPublishHooks({
  projectHandle: 'your-awesome-project',
  channelHandle: 'some-channel',
  publishHook: functionA,
  unpublishHook: () => {}
})
…
.registerPublishHooks({
  projectHandle: 'your-awesome-project',
  channelHandle: 'some-channel',
  publishHook: () => {}, // this line
  unpublishHook: functionB
})
```

?

Same question if instead of `() => {} // this line` you put `null`, `undefined`, `false`, `20`?

I think we have many valid options here, I don't see an obvious one. I would rather have an API that behaves the way I expect it to behave.

peyerluk: I would just register the hooks which are specified. And encourage users to do it all at once for one channel. In the end I think this leads to better code.

Registering them all at once also indicates in a better way that this is more like a setup than registering an event listener. Because we can have only one publish and unpublish hook per channel. It is not possible e.g. to register two publish hooks for the same channel.

vhf: It works for me. I'll explicitly check for `undefined` then.

vhf:
@peyerluk Please review.

reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,88 @@ copy: [
...
+Since hooks get passed the document `documentType`, it is up to the hook implementer to filter by `documentType` if need be. This can be used to apply different modifications to documents that have different `documentTypes`.
+
+Examples for each hook type:
+
+### Publish/Unpublish Hooks
+
+These two are set at once on the `documents` feature.
+
+```js
+liServer.features.api('li-documents').registerPublicationhHook({
```
peyerluk: Shouldn it be plural? E.g. 'registerPublicationhHooks'?

reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,88 @@ copy: [
 ```
 
 
+## Hooks: before render, on publish, on unpublish
+
+A few APIs are provided to allow hooking into the document (un)publication process and into the rendering pipeline.
+
+Although these hooks should preferably be registered before the server gets initialized, it is also possible to register them at runtime using the same APIs. (This is particularly handy for testing purpose, but also useful if you create projects or channels at runtime and need to set hooks for these.)
+
+All hooks are asynchronous functions that are registered synchronously using the common signature: `({projectHandle, channelHandle, hook})`. The function `hook` passed here has the signature `({documentType, payload}, callback)`. The `payload` is the only thing that vary by hook type.
```
peyerluk: This isn't accurate anymore.

reference-docs/server-configuration/channel-config.md
vhf: Yep, and there's an extra `h` in there!

reference-docs/server-configuration/channel-config.md
```diff
@@ -142,16 +142,16 @@ copy: [
 
 A few APIs are provided to allow hooking into the document (un)publication process and into the rendering pipeline.
 
-Although these hooks should preferably be registered before the server gets initialized, it is also possible to register them at runtime using the same APIs. (This is particularly handy for testing purpose, but also useful if you create projects or channels at runtime and need to set hooks for these.)
+Although these hooks should preferably be registered before the server gets initialized (using [Server Initialized Hooks](../../reference-docs/server-extensions/server-initalization.md#initialized-hooks)), it is also possible to register them at runtime using the same APIs. (This is particularly handy for testing purpose, but also useful if you create projects or channels at runtime and need to set hooks for these.)
```
vhf: This link leads to the new section of that file: https://github.com/upfrontIO/livingdocs-server/pull/1589

reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,88 @@ copy: [
...
+
+Each hook is tied to a specific channel of a specific project. For a single channel, you can only register one hook of each type:
+
+```js
+registerPublicationHooks({projectHandle: 'foo', channelHandle: 'bar', publishHook: myHook})
+// the following register call will overwrite `myHook` with `myOtherHook`:
+registerPublicationHooks({projectHandle: 'foo', channelHandle: 'bar', publishHook: myOtherHook})
+```
+
+Since hooks get passed the document `documentType`, it is up to the hook implementer to filter by `documentType` if need be. This can be used to apply different modifications to documents that have different `documentTypes`.
```
peyerluk: Since hooks get passed the `documentType` of a document, ....

reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,88 @@ copy: [
 ```
 
 
+## Hooks: before render, on publish, on unpublish
+
+A few APIs are provided to allow hooking into the document (un)publication process and into the rendering pipeline.
+
+Although these hooks should preferably be registered before the server gets initialized (using [Server Initialized Hooks](../../reference-docs/server-extensions/server-initalization.md#initialized-hooks)), it is also possible to register them at runtime using the same APIs. (This is particularly handy for testing purpose, but also useful if you create projects or channels at runtime and need to set hooks for these.)
+
+All hooks are asynchronous functions, registered asynchronously using the common signature: `({projectHandle, channelHandle, hooks}, callback)`. The function `hook` passed here has the signature `({documentType, payload}, callback)`. The `payload` is the only thing that vary by hook type.
```
peyerluk: Some corrections:

All hooks are asynchronous functions, registered asynchronously using the common signature: `({projectHandle, channelHandle, namedHook}, callback)`. The hook provided must have the signature `({documentType, payload}, callback)`. The `payload` is the only thing that varies by hook type.

reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,88 @@ copy: [
...
+    payload // payload is {documentVersion, renditions}
+  }, callback) => {
+    liServer.log.info(`Hook called for documentType: ${documentType}!`)
+    liServer.log.debug({
+      documentVersion: payload.documentVersion,
+      renditions: payload.renditions
+    })
+    callback()
+  },
+  unpublishHook: ({documentType, payload}, callback) => {callback()}
```
peyerluk: We have to provide the payload of the unpublish hook as well. Maybe we should describe the payload of all hooks also outside of the example. As your write above the payload is the thing that varies so it must be described for each hook.

reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,88 @@ copy: [
...
+This one hooks into the `render-pipeline` feature. Here is a full example including server initialization:
+
+```js
+const server = require('livingdocs-server/core')
+const config = require('livingdocs-server/conf')
+const liServer = server(config)
+const port = liServer.config.get('server:port')
+
+
+liServer.registerInitializedHook((done) => {
```
peyerluk: You don't do that in the other example. Make it consistent.

reference-docs/server-configuration/channel-config.md
```diff
@@ -157,6 +138,88 @@ copy: [
...
+  liServer.features.api('li-render-pipeline').registerRenderHook({
```
peyerluk: I would also do `registerRenderHooks` here to make it consistent. Plus we don't have to change the name if we add hooks.

reference-docs/server-configuration/channel-config.md
vhf: Since the doc says these hooks can be registered both at runtime and using server hooks, I kept the first example simple and introduced this one with a sentence clearly stating the difference with the previous example:

> Here is a full example including server initialization:

Would you rather drop the full example or have all examples use the same boilerplate?

peyerluk: Let's do the full example everywhere. A lot of people don't read documents from top to bottom and then it can cause confusion if people just copy an example they need.

reference-docs/server-configuration/channel-config.md
vhf: Done.

reference-docs/server-configuration/channel-config.md
vhf: Done.

reference-docs/server-configuration/channel-config.md
vhf: Done.

reference-docs/server-configuration/channel-config.md
vhf: Done.

reference-docs/server-configuration/channel-config.md
vhf: Fixed.

peyerluk:
@vhf I just realized we have to add the `document-lists` hooks as well.

vhf:
@peyerluk Document lists hooks added in the last commit.

Up for review.

peyerluk:
Can you open a release-branch and pr for the September release (like https://github.com/upfrontIO/livingdocs/pull/133) and merge this branch in there?

vhf:
Done.

PR #124, opened 2017-08-29
docs(hooks): Document server initialized hooks

vhf:


reference-docs/server-extensions/server-initalization.md
```diff
@@ -17,3 +17,15 @@ liServer.listen(port, function (err) {
   console.log('Listening on http://localhost:%s', port)
 })
 ```
+
+## Initialized Hooks
+
+Hooks to be executed right after the server initialized can be registered on a Livingdocs Server instance.
```
peyerluk: We should mention that they are executed before the `listen()` function (before the server starts to listen to a port).

vhf: Yep, I'll look into it. I'll try to detail the startup steps and where these hooks get executed.

vhf:
@peyerluk I wasn't sure how to explain that the hooks are executed before `listen` since `listen` calls `initialize`.

peyerluk:
@vhf I rewrote it a bit to clarify.

peyerluk:
Merged into: https://github.com/upfrontIO/livingdocs/pull/133

PR #125, opened 2017-08-29
chore(docs): Fix a few typos

vhf:


concepts/glossary.md
```diff
@@ -5,7 +5,7 @@
 - Livingdoc
 
   A representation of content. It separates content from structure by using
-  `Components`. A `Livingdoc` references a specific `Design` wich defines the
+  `Components`. A `Livingdoc` references a specific `Design` which defines the
```
vhf: @gabrielhase We have two glossaries and it seems their content mostly overlap.

gabrielhase: yes, we can delete the one that is not all upercase

vhf: Added a commit deleting it.

PR #127, opened 2017-08-31
chore(docs): Remove nbsp in Markdown titles

vhf:
Markdown headings are specified as `#<space+><name>`. This PR changes all `#<nspb><name>` to replace the nbsp with a space in order to have these parsed as titles instead of literal `# foo`.


vhf:
Swiss-german then I guess, I never had this. 🦊

PR #131, opened 2017-09-01
chore(release): better explained steps and improved wording

rebelact:


reference-docs/maintenance/release-management/how.md
```diff
@@ -7,22 +7,33 @@
 
 ### Assumption
 
-The release use `v2.1.0` as a base tag.
+We want a release with semantic version `v2.1.0`.
```
peyerluk: Please add some clarification. Is `v2.1.0` the release name? Does it belong to the `livingdocs-server` or `livingdocs-editor` or neither?

reference-docs/maintenance/release-management/how.md
```diff
@@ -7,22 +7,33 @@
...
 
 ### Steps
 
-#### Optional
-**If** `v2.1.0` is not the latest tag/release and to avoid potential versioning conflicts one need to execute:
+Execute this for the `livingdocs-server` and `livingdocs-editor`.
```
peyerluk: Maybe we can create what our goal is here.

Like:

> We want to create a release-branch for both the `livingdocs-server` and the `livingdocs-editor`. Execute the following steps in each of those two repositories.

reference-docs/maintenance/release-management/how.md
```diff
@@ -7,22 +7,33 @@
...
+
+#### If `v2.1.0` already exists
```
peyerluk: This «if `v2.1.0` already exists» structure doesn't really work if there are two repos with different versions. Or do I understand it wrong?

peyerluk: We would have to phrase it differently then.

Something along the lines of:

> Let's go through creating a release-branch for the livingdocs-server as an example.

peyerluk: /cc @mthpvg

reference-docs/maintenance/release-management/how.md
```diff
@@ -7,22 +7,33 @@
...
+**Please note** that the release process does not create the semantic version you want to release.
+
 #### Mandatory
 
-1. `./node_modules/@livingdocs/release-tools/li-release create-maintenance-release --base-tag 2.1.0 --npm-auth-token <token>`
+**Prerequisites**: `v2.1.0` should already exists as a release in the upstream repository. If the release has been executed properly the npm package with version `v2.1.0` should also be uploaded to `npm`.
+
+The line below creates a new branch `maintenance-2.1.x`, where you can make patches for `2.1.x`. It won't conflict with 2.2.0 and newer versions. It is where hotfixes will be released.
+
+1. `./node_modules/@livingdocs/release-tools/li-release create-maintenance-release --base-tag 2.1.x --npm-auth-token <token>`
```
DaRaFF: `--base-tag 2.1.x` is wrong. It always must be a concrete existing version. In this case it's `--base-tag 2.1.0`

reference-docs/maintenance/release-management/how.md
DaRaFF: I think without example images this whole process is hard to understand.

DaRaFF:
This PR is not used anymore because of https://github.com/upfrontIO/team/pull/59

PR #133, opened 2017-09-01
August Release (maintenance-2017-08)

peyerluk:
Documentation updates for the august release.


peyerluk:
@rebelact Here are the updated docs for the new release.

PR #136, opened 2017-09-15
Update Documentation Regarding the `grunt setup` fix for Docker Containers

DaRaFF:
Pull Request: https://github.com/upfrontIO/livingdocs-server/pull/1617

### DIFF
--- a/reference-docs/server-configuration/config.md
+++ b/reference-docs/server-configuration/config.md
@@ -33,7 +33,14 @@ db: {
   database: 'li_local',
   pool_min: 1,
   pool_max: 5,
-  ssl: false
+  ssl: false,
+  // During the initial db setup e.g. with 'grunt setup',
+  // this parameter controls whether the database setup should run in a dedicated docker
+  // container. That way you don't need to install the postgres `createdb` and `dropdb` commands.
+  // `grunt setup` is not supported, when you start docker containers with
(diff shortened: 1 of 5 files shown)


reference-docs/server-configuration/config.md
```diff
@@ -33,7 +33,13 @@ db: {
   database: 'li_local',
   pool_min: 1,
   pool_max: 5,
-  ssl: false
+  ssl: false,
+  // During the initial db setup f.e. with 'grunt setup',
```
marcbachmann: There's no `f.e.` in english. ;)
Use `e.g.`. Indeed weird.

reference-docs/server-configuration/config.md
```diff
@@ -33,7 +33,13 @@ db: {
...
+  // this parameter controls if a docker container for the database setup commands is used.
```
marcbachmann: >  if a docker container for the database setup commands is used.

whether the database setup should run in a dedicated docker container. That way you don't need to install the postgres `createdb` and `dropdb` commands.

reference-docs/server-configuration/config.md
```diff
@@ -33,7 +33,13 @@ db: {
...
+  // default = true
```
marcbachmann: Please move below the other comments

PR #138, opened 2017-09-29
Documentation for doc-includes

gabrielhase:
Documentation for the `doc-include` feature.


gabrielhase:
@DaRaFF I chose you as an "unspoiled" reviewer, read through it and see if you can understand it from the documentation
@marcbachmann you know the stuff inside out, please check if I made any mistakes

I would recommend looking at it directly in the gitbook preview. Just do `gitbook serve` on the command line.

reference-docs/doc-includes/editor_customization.md
```diff
@@ -0,0 +1,205 @@
...
+}
+```
+
+- `directive`, almost the same as in the sidebar component (see above), but this time it is only passed in as a one-way binding so if you want to change parameters on the directive you need to use the `dispatch` method below
+- `dispatch`, you usually call this method once you want to close your modal, e.g. in the case of an article embed, when the user has selected an article to embed. You can call the method as an update and as a cancel: `this.dispatch({action: 'update', params: {foo: 'bar'}})` will close the modal and update the parameter `foo` on the `doc-include` to the value `'bar'`. `this.dispatch({action: 'close'})` will simply close the modal without updating any parameters of the `doc-include`.
+
+### An example
+
+`controller.js`
+```controller.js
```
marcbachmann: Weird language annotation. Please remove `controller.` here.

reference-docs/doc-includes/editor_customization.md
```diff
@@ -0,0 +1,205 @@
...
+const _get = require('lodash/get')
+const BrightcoveVideo = require('./model')
+
+module.exports = class BrightcoveVideoController {
+
+  static get $inject () { return ['editor'] }
```
marcbachmann: I don't like to have an example that exposes internals like `editor`.
It would be nice if we could pass the document differently.

reference-docs/doc-includes/editor_customization.md
```diff
@@ -0,0 +1,205 @@
...
+      id: this.id,
+      language: this.language
+    })
+  }
+}
+```
+
+The model is optional (you could also write all code in the controller but we advise the use of models). There are 2 important things to note in the model:
+- `directive.getParams()`, gets the current parameters (including `defaultParams`) of the selected `doc-include`
+- `directive.addParams({foo: 'bar'})`, saves the given parameters to the existing ones, overwrites keys if duplicate, NOTE: this does not overwrite the whole params object, only when a key collides
```
marcbachmann: The `addParams` results in an unexpected state as it doesn't override existing parameters.
Mathieu also ran into a similar issue in here: https://github.com/upfrontIO/livingdocs-planning/issues/817
I think I've mentioned that in some issue in the framework that the method tries to do too much. IMO we should simplify it by exposing a `setParams` method in the framework.

reference-docs/doc-includes/editor_customization.md
```diff
@@ -0,0 +1,205 @@
...
+        </form>
+      </div>
+    </div>
+  </div>
+</div>
+```
+
+Last but not least you have an angular component template. This will be rendered in the sidebar. Regarding HTML structure and classes see the [Livingdocs styleguide](https://beta.livingdocs.io/styleguide.html).
+
+### An example with dispatch
```
marcbachmann: In general I think those examples are too complex, they require too much knowledge of the setup.
In long-term we should simplify the api.

reference-docs/doc-includes/embed_and_list.md
```diff
@@ -0,0 +1,180 @@
...
+Lets see an example component in a Livingdocs design.
+
+```
+<script type="ld-conf">
+{
+  "name": "gallery-embed",
+  "label": "Gallery embed",
+  "directives": {
+    "gallery-embed": {
+      "service": "embed-teaser",
```
marcbachmann: Didn't we prefix that with `li-`?

reference-docs/doc-includes/embed_and_list.md
```diff
@@ -0,0 +1,180 @@
...
+    output:
+      'json':
+        outputRenderer: appRenderer
+```
+
+There are 2 renditions here, web and app. For the web, the `embed-teaser` include is resolved (`resolveIncludes`) thus in my published HTML documents I will see the rendered `doc-include` according to the rendering function I wrote. For the app, no includes are resolved, thus I will only get placeholders containing the parameters without any rendering happening.
+
+### List
+
+![List](./list.png)
```
marcbachmann: The ui of that example is still broken. We should either remove or improve it.

reference-docs/doc-includes/editor_customization.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+The only thing that differs from the prior example is the use of the `dispatch` method in `select`.
+
+### Using an iframe
+
+In the [server configuration](./server_customization.md#include-ui-options) we learned that you can also register an external iframe as a user interface in a modal. We will not show any example code for this because you can use whatever code you like in the iframe (doesn't need to be angular).
+One thing to note though is how to close or update the Livingdocs editor modal from inside your custom iframe. The Livingdocs modal listens to `postMessage` for updates. So in your code you can do:
+```
+window.top.postMessage({action: 'update', params: {"someParam": "test"}})
```
marcbachmann: I thought postMessage requires the `targetOrigin` parameter for security reasons. https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

marcbachmann: We should document how we pass initial arguments to the iframe. I thought we're passing a query string to the url of the iframe. Maybe there's also an event.

reference-docs/doc-includes/embed_and_list.md
```diff
@@ -0,0 +1,180 @@
+## Core doc-includes
+
+The Livingdocs core predefines 2 editor user interfaces for the `doc-include` services: `embed-teaser` and `list`.
```
marcbachmann: We should rename `embed-teaser` as this is conflicting with the embed functionality we have in the editor.

reference-docs/doc-includes/editor_customization.md
gabrielhase: Agreed. I'll make a simpler example.

reference-docs/doc-includes/editor_customization.md
gabrielhase: Mathieus problem was unrelated. It was a problem how he passed references in the editor once he had fixed that the framework worked as expected.
I agree thought that `addParams` is a bit weird. @peyerluk what is the reasoning for this choice?
In any case, this PR is about documentation. Implementation issues should be created and discussed in the planning repo.

reference-docs/doc-includes/embed_and_list.md
gabrielhase: Only the angular components to my knowledge

reference-docs/doc-includes/embed_and_list.md
gabrielhase: Don't remove it! Bluewin and SZ are using this and are relying on it. Also they are reasonably happy with it. If you want to improve something please discuss / coordinate with me and Genc beforehand.

reference-docs/doc-includes/editor_customization.md
gabrielhase: I pretty much copied the documentation from your editor PR: https://github.com/upfrontIO/livingdocs-editor/pull/1509
So this is not working?

reference-docs/doc-includes/embed_and_list.md
gabrielhase: I agree that `embed-teaser` is a very poor naming choice. The problem here is that this is a breaking change for all of our customers and requires a full data migration... If you find a way to sell them the effort for a rename, I'd be happy :)

reference-docs/doc-includes/editor_customization.md
marcbachmann: No, it is. I just wasn’t sure about the api anymore.

reference-docs/doc-includes/embed_and_list.md
```diff
@@ -0,0 +1,180 @@
...
+The `embed-teaser` is an article embed. The user can click the component containing the `embed-teaser` include and the sidebar shows a user interface that lets the user select an article to be embedded as a teaser.
+The `list` represents a list of articles, commonly seen on startpages of newspapers. It assumes a manually sorted list that can be created with the Livingdocs list interface (see main navigation inside the Livingdocs editor). The sidebar renders a user interface where a user can select a list or re-order the articles in that list.
```
DaRaFF: The term 'Livingdocs list interface' is not clear for me. (Because I know the editor, I know what you mean, but I think a reader wouldn't understand that).

marcbachmann: 👍  We should create cross references to explain such details.

reference-docs/doc-includes/server_customization.md
```diff
@@ -0,0 +1,217 @@
...
+This section explains how you setup the server to render `doc-include`s and gives an exhaustive listing of all options. It does not explain how to create custom editor user interfaces, this will be explained [here](./editor_customization.md).
+
+### Resolve options
+
+Includes can be resolved in two modes when rendering a document. Either a `<ld-include>` tag can be rendered so the include can be resolved outside of Livingdocs (e.g by a script in the browser or a server that delivers the rendered HTML to a browser). Or the includes can be resolved on the Livingdocs server in the rendering of the document.
+
+#### Unresolved include
+
+An unresolved include looks like this:
+```html
```
DaRaFF: I would mention that this is a design component.

reference-docs/doc-includes/embed_and_list.md
```diff
@@ -0,0 +1,180 @@
...
+        outputRenderer: appRenderer
+```
+
+There are 2 renditions here, web and app. For the web, the `embed-teaser` include is resolved (`resolveIncludes`) thus in my published HTML documents I will see the rendered `doc-include` according to the rendering function I wrote. For the app, no includes are resolved, thus I will only get placeholders containing the parameters without any rendering happening.
+
+### List
+
+![List](./list.png)
+
+The image above shows how a manually sorted list of article teasers is created in Livingdocs. Upon selecting the component, the user can select a list in the sidebar ("bar" in the example). If the user presses "edit" in the sidebar, the user gets a second sidebar (to the left) where she can change the order of articles and add/remove articles. When she presses "Publish List" the preview is updated.
```
DaRaFF: In my view it's very dangerous to provide this list doc-include with the current implementation. If an editor is not aware, that he is maintaining a *real* list here, then it's possible that he modifies this lists without knowing the consequences e.g. for lists or pages which this list is referenced to.

reference-docs/doc-includes/intro.md
```diff
@@ -0,0 +1,80 @@
+## Doc-Includes
```
DaRaFF: Some visualisation of the dependencies between all parts of a successful integration of a doc-include could support the comprehensibility for the user.

reference-docs/doc-includes/server_customization.md
```diff
@@ -0,0 +1,217 @@
+## Doc-include server customizations
+
+This section explains how you setup the server to render `doc-include`s and gives an exhaustive listing of all options. It does not explain how to create custom editor user interfaces, this will be explained [here](./editor_customization.md).
+
+### Resolve options
+
+Includes can be resolved in two modes when rendering a document. Either a `<ld-include>` tag can be rendered so the include can be resolved outside of Livingdocs (e.g by a script in the browser or a server that delivers the rendered HTML to a browser). Or the includes can be resolved on the Livingdocs server in the rendering of the document.
```
DaRaFF: This point is not fully clear for me.
- how can i resolve an ld-include in the browser? Is this the same resolving as it is done in the editor?
- when you are talking about the resolving on the server. Do you talk of the rendering in the renditions? Or do you also talk of the resolving when a document will be delivered to the editor for the preview?

reference-docs/doc-includes/server_customization.md
```diff
@@ -0,0 +1,217 @@
...
+}
+```
+
+The `resolveIncludes` array lists all includes that should be resolved inside of Livingdocs. You can configure this by channel and by rendition.
+
+### Registration of an include renderer in the server
+
+You can register an include renderer in the Livingdocs Server using the API of the `li-includes` feature.
+
+The include is rendered for two cases:
```
DaRaFF: Here is the rendering explained much better than at the beginning (it was puzzling for me).

Please make the different rendering options more obvious (you have explained it in all the different pages a few times, but always a little bit different).

1. You can render the doc-include only for the editor in editing mode
2. You can render the doc-include only for a publication via public API
3. You can combine option 1 + 2

reference-docs/doc-includes/embed_and_list.md
```diff
@@ -0,0 +1,180 @@
+## Core doc-includes
```
DaRaFF: Immediately after the intro I was reading this page 'embed and list' and it was hard to understand for me. It should be more obvious that this is an integration example and that the basics should be already clear like the server and editor integration.

reference-docs/doc-includes/embed_and_list.md
gabrielhase: The list tool is a question all for itself... But we have it, all clients use it productively, i.e. it should be documented

reference-docs/doc-includes/server_customization.md
gabrielhase: It isn't :) This is the output of the render pipeline if you don't resolve includes

reference-docs/doc-includes/server_customization.md
gabrielhase: The browser is a digression here and seems to complicated. I just remove that note.

gabrielhase:
@DaRaFF please check again, I reworked the introduction.

reference-docs/doc-includes/server_customization.md
```diff
@@ -4,11 +4,11 @@ This section explains how you setup the server to render `doc-include`s and give
 
 ### Resolve options
 
-Includes can be resolved in two modes when rendering a document. Either a `<ld-include>` tag can be rendered so the include can be resolved outside of Livingdocs (e.g by a script in the browser or a server that delivers the rendered HTML to a browser). Or the includes can be resolved on the Livingdocs server in the rendering of the document.
+Includes can be resolved in two modes when rendering (i.e. publishing) a document. Either a `<ld-include>` tag can be rendered so the include can be resolved outside of Livingdocs (e.g. in a microservice). Or the includes can be resolved on the Livingdocs server in the rendering of the document.
```
DaRaFF: much better

PR #142, opened 2017-10-14
Add return statement to metadata plugin examples

gabrielhase:
@DaRaFF fyi people tend to just copy sample code so the sample code should never "break" stuff. We had the problem with somebody just copying the code and then wondering why all metadata fields became `null` after saving.


DaRaFF:
thank you @gabrielhase !

PR #143, opened 2017-10-19
Reference and Guides

gabrielhase:
Introduce a reference section and clearly mark design and customizing as guides.


gabrielhase:
@peyerluk I tried to give some heed to our discussion. What do you think?

PR #144, opened 2017-10-23
Add up/downstream term to the glossary

DaRaFF:
* Add up/downstream term to the glossary
* Introduce alphabetical order

### DIFF
--- a/DICTIONARY.md
+++ b/DICTIONARY.md
@@ -35,4 +35,4 @@ Represents an organisation and consists of users and channels.
 
 ## Upstream
 
-*Upstream* is a package/repository which is integrated in another project called *downstream*. In our case *upstream* means our core editor and server is provided as a npm package (semantic versioning) and can be used by a *downstream* project e.g. `livingdocs-server-downstream`.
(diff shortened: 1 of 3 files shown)


GLOSSARY.md
```diff
@@ -1,29 +1,38 @@
...
 
 A print publishing system.
+
+## Project
+
+Represents an organisation and consists of users and channels.
+
+## Upstream
+
+*Upstream* is a package/repository which is integrated in another project called *downstream*. As an example: There is an *upstream* project called livingdocs-server(core) which is integrated/used by a *downstream* project livingdocs-server-downstream.
```
gabrielhase: That sounds too generic. We don't need to explain what an upstream in general is but only in the context of our software, so please tell that it is our core software, that it is an npm package and that we use semantic versioning.

marcbachmann: I'd keep the first sentence, replace the `As an example` with `In our case` or something like that.

DaRaFF:
@gabrielhase please check it again. I renamed glossary to dictionary.

DICTIONARY.md
```diff
@@ -1,29 +1,38 @@
...
 
 A print publishing system.
+
+## Project
+
+Represents an organisation and consists of users and channels.
+
+## Upstream
+
+*Upstream* is a package/repository which is integrated in another project called *downstream*. In our case *upstream* means our core editor and server is provided as a npm package (semantic versioning) and can be used by a *downstream* project e.g. `livingdocs-server-downstream`.
```
gabrielhase: I am nitpicking :) please add a hint that the framework does not have a downstream/upstream separation and can not be installed separately.

PR #148, opened 2017-11-04
SSO guide

gabrielhase:
boilerplate:
- https://github.com/upfrontIO/livingdocs-server-boilerplate/pull/73
- https://github.com/upfrontIO/livingdocs-editor-boilerplate/pull/54

### DIFF
--- a/reference-docs/server-configuration/metadata.md
+++ b/reference-docs/server-configuration/metadata.md
@@ -1,6 +1,3 @@
-DEPRECATED: The feature is not working at the moment and needs to be integrated into the public API. Then this needs to be re-documented
-
-
 # Metadata on the Server
 
 ## Scope
--- a/reference-docs/server-configuration/metadata.md
+++ b/reference-docs/server-configuration/metadata.md
@@ -1,3 +1,6 @@
+DEPRECATED: The feature is not working at the moment and needs to be integrated into the public API. Then this needs to be re-documented
+
+
 # Metadata on the Server
 
 ## Scope
--- a/walkthroughs/github-login.md
+++ b/walkthroughs/github-login.md
@@ -1,0 +1,105 @@
+## SSO example with Github login
+
+The boilerplate server and editor provide an example for a SSO login using Github as an authentication provider.
+The bulk of the implementation is in the server. Check out the `app/sso` directory in the [boilerplate server](https://github.com/upfrontIO/livingdocs-server-boilerplate).
+
+### How the authentication flow works
+
+As you can see in the boilerplate SSO example, we use `passport-light` (a module by Marc Bachmann) and the `passport-github` strategy. Passport has 300+ strategies on its [website](http://www.passportjs.org/) so chances are that yours exists already and you can pretty much re-use the code from the boilerplate example.
+If for some reason you can not use passport, this is also not a problem. The flow for authentication is pretty simple:
+1. The Livingdocs editor makes a call to the Livingdocs server to ask for authentication (you can configure the route)
+2. The Livingdocs server redirects this call to the third-party SSO provider
+3. The user logs in to the third-party SSO provider (using the user interface of the SSO provider, e.g. Github login)
+4. The third-party SSO provider redirects back to the Livingdocs server passing information about the logged in user (in the minimum a unique id)
+5. The Livingdocs server checks the identity table for a user with the external id passed in by the SSO provider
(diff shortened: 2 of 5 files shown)


gabrielhase:
@DaRaFF please review.
I guess you don't know how SSO works yet, so please review the documentation and example implementation in the boilerplate and see if you can understand it like this.

reference-docs/server-configuration/metadata.md
```diff
@@ -1,3 +1,6 @@
+DEPRECATED: The feature is not working at the moment and needs to be integrated into the public API. Then this needs to be re-documented
```
DaRaFF: What exactly is not working? If metadata on the server wouldn't work we would have a huge problem.

walkthroughs/github-login.md
```diff
@@ -0,0 +1,105 @@
...
+If for some reason you can not use passport, this is also not a problem. The flow for authentication is pretty simple:
+1. The Livingdocs editor makes a call to the Livingdocs server to ask for authentication (you can configure the route)
+2. The Livingdocs server redirects this call to the third-party SSO provider
+3. The user logs in to the third-party SSO provider (using the user interface of the SSO provider, e.g. Github login)
+4. The third-party SSO provider redirects back to the Livingdocs server passing information about the logged in user (in the minimum a unique id)
+5. The Livingdocs server checks the identity table for a user with the external id passed in by the SSO provider
+6. The Livingdocs server creates a Livingdocs access token for the user found in (5)
+7. The Livingdocs server redirects to `${editorHost}/login#token=${accessToken}` to login the user
```
DaRaFF: I like the description, because one can understand what happens under the hood.

walkthroughs/github-login.md
```diff
@@ -0,0 +1,105 @@
...
+  providers: [{
+    id: 'github',
+    strategy: 'link',
+    label: 'Log in via Github',
+    url: 'http://localhost:9090/auth/github'
+  }]
+}
+```
+
+This will render a button on the login screen, reading "Log in via Github" and request the authentication from `/auth/github`. Your Livingdocs server needs to implement this route with your SSO implementation.
```
DaRaFF: It would be nice to have a default implementation already on the upstream server for github authentication, so the customers can only use it and don't have to implement it on the downstream server.

walkthroughs/github-login.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Implementing an external login provider on the server
+
+*Note: the following walkthrough (and the boilerplate example) makes use of Livingdocs core code APIs. Those APIs can not assumed to be stable so you will likely need to change your code as we introduce breaking changes in the future. If you closely watch the release notes before each update this shouldn't be a big issue though.*
```
DaRaFF: I'm not sure if this comment is necessary, even when it's well intentioned. Everyone knows that an API is an interface and that API can change. Otherwise you have to add this note to every walkthrough document.

walkthroughs/github-login.md
```diff
@@ -0,0 +1,105 @@
...
+const projectApi = server.features.api('li-projects')
+```
+
+You will also need to define the id of the project where your users will be created in. The boilerplate example creates users in a common project, of course you could also create a new project for each user.
+Most likely you will also need to get some configured access tokens for your third-party SSO, in the example of github this is an app id and secret. If you want to try the Github example you need to create an oAuth app in your Github account and copy the id and secret from there.
+
+### routes
+
+You will need to define the route for authentication that you configured in the editor before.
+```routes.js
```
DaRaFF: \```javascript
// routes.js
 module.exports = {
   endpoints: [{
     method: 'get',
     path: '/auth/github',
     title: 'Create a session using github',
     action: 'authenticate'
   }]
 }
\```

reference-docs/server-configuration/metadata.md
gabrielhase: damn, wrong file. thx for pointing that out

walkthroughs/github-login.md
gabrielhase: yes. I just don't want to touch the core server too much :) but @marcbachmann knows about this

walkthroughs/github-login.md
gabrielhase: Actually, not. This one uses "internal" APIs while the other guides only use the "ticked off by Marc" APIs ;)

PR #151, opened 2017-11-22
doc: image services configuration

mthpvg:
### Planning
https://github.com/upfrontIO/livingdocs-planning/issues/1532

@gabrielhase can you review?

### DIFF
--- a/concepts/images/image-services.md
+++ b/concepts/images/image-services.md
@@ -66,7 +66,7 @@ Srcset only works for inline images (`img` tag). If you use a background image, 
 
 ## Configuring an image service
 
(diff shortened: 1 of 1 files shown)


concepts/images/image-services.md
```diff
@@ -78,38 +78,11 @@ The ImgIX configuration uses `srcset`. If you don't know about this HTML standar
...
-        widths: [
-          2048,
-          1024,
-          620,
-          320
-        ],
-        sizes: ['100vw']
-      }
-    }
+    imageService: 'imgix'
```
marcbachmann: That is still needed?

mthpvg: I saw that too late... I'm not the best judge for that, but it allows a single server to answer to multiple editors with different image services.

gabrielhase: yes, the selection of the concrete image service is still in the editor. all the configuration is in the server. and yes, the server can have configurations for multiple image services if that is liked.

PR #159, opened 2018-01-31
Add Design loader feature configuration

peyerluk:


reference-docs/server-configuration/config.md
```diff
@@ -193,9 +193,37 @@ emails: {
 }
 ```
 
+#### Design Loader
 
-#### Designs
+The DesignLoader offers different options how designs are loaded. By default designs
+are loaded from the local database.
+
+Options include:
```
mthpvg: What do you think of:

The three options include:
1. Preload a design into memory (useful for testing, or if you only change designs during a deploy) **default**
  ```js
  designLoader: {
    localDesigns: [{ // optional
      path: '/designs/timeline/v1.1.0' // path to design
    }],
    cacheSize: 100 // optional, defaults to '20'
  }
  ```
2. Load a specific design from a url
  ```js
  designLoader: {
    hostedDesigns: [{ // optional
      designName: 'timeline',

      // The final request for timeline design v1.1.0 will go to this url:
      // -> http://assets.livingdocs.io/timeline/1.1.0
      url: 'http://assets.livingdocs.io/timeline'
    }],
    cacheSize: 100 // optional, defaults to '20'
  }
  ```
3. Load a design from another Livingdocs Server
  ```js
  designLoader: {
    designRepository: { // optional, defaults to the local design server
      remoteHost: 'http://api.livingdocs.io'
    },
    cacheSize: 100 // optional, defaults to '20'
  }
  ```

It could emphasize that those options are mutually exclusive.

One question: In those three scenarios the server will fetch the design during the setup, cache it and server it through the new endpoint you created?

peyerluk:
> ... but what happens when all 3 options are configured? Is there a preferred one?

They all work at the same time.

Algorithm:

1. check if for this specific design a custom url is configured
2. check if the design is preloaded
3. load from remote design server if configured otherwise load from local server

reference-docs/server-configuration/config.md
peyerluk: I'll have to clarify the docs. They all work at the same time. The designLoader is shared between all projects on a server. So it can deal with many designs at the same time. And there can be different strategies for different designs.

peyerluk: > In those three scenarios the server will fetch the design during the setup

No the server only fetches a design when requested (except the preloaded designs). In the case of the hostedDesigns it cannot prefetch anything  as it does not know which version to ask for.

peyerluk:
@DaRaFF @mthpvg I added some clarifications. Can you check if it is easier to understand now?

mthpvg:
Yes, now I get it. Thx.

DaRaFF:
Sorry if I'm a little bit tedious, but I still don't fully understand the description (I didn't check the code).

> 1. check if the design is preloaded
> 2. check if for this design a custom url is configured. If yes it will append the version to this url and fetch from that url. If this fails it will not continue to step 3.
> 3. load from a remote design server if a designRepository is configured otherwise load from local server

- What is a custom url? Is this if you add something either in `hostedDesigns` or `localDesigns`? What happens if `hostedDesigns` and `localDesigns` have the same entries, which one is preferred?
- Why do we have `hostedDesigns` and `designRepository`, is this not the same (you load something from a remote host)?

> path: '/designs/timeline/v1.1.0' // path to design

- path to design means a filepath, right?

peyerluk:
@DaRaFF I updated the docs again. Please recheck :)

DaRaFF:
@peyerluk I think i got it except one small thing. A local design `path: '/designs/timeline/v1.1.0' // path to design` is a design on a local running design server? Is it somehow possible to load a design from a file (would be nice for the development)

marcbachmann:
> A local design path: '/designs/timeline/v1.1.0' // path to design is a design on a local running design server?

The local designs are files from disk, not from the design server. `hostedDesigns` are the ones from a server. I hope this whole api is only "temporary" until we have the new one.

PR #167, opened 2018-02-27
WIP: Add reference-docs about storage configs

peyerluk:
Document the `storage` option available in features that store static files. And document how S3 can be configured with a proxy.

## Todo

- Check if it is a problem if a user sets `httpOptions.agent` in a `storage` config and also the environment variable `HTTP_PROXY` (in which case `httpOptions.proxy` will be set as well).
https://github.com/upfrontIO/livingdocs-server/blame/validate-image-service-config-master/lib/blob-store/strategies/s3.js#L9
- Check if we can find a solution where customers who use a proxy must not require `proxy-agent` in their configuration (this can prevent customers from using JSON files for their configuration). 
- Update the server config reference-docs so that all features that can use the storage config mention it properly

### DIFF
--- a/reference-docs/server-configuration/file-storage-config.md
+++ b/reference-docs/server-configuration/file-storage-config.md
@@ -1,0 +1,56 @@
+# Storage Configuration
+
+Multiple features need to store static files. For this they offer a `storage`
+option that allows for different storage strategies.
+
+
+## Amazon S3 strategy
+
+Example Amazon S3 configuration to store images:
+```js
+images: {
+  storage: {
+    strategy: 's3',
+    config: {
+      bucket: 'livingdocs-images-dev',
+      region: 'us-west-1',
+      accessKeyId: process.env.ENV_VAR,
+      secretAccessKey: process.env.ENV_VAR,
+    }
+  }
+}
+```
+
+
+#### Using a proxy with S3
+
+For S3 it is necessary to use the proxy-agent module. For more info see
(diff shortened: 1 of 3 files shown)


reference-docs/server-configuration/file-storage-config.md
```diff
@@ -0,0 +1,56 @@
...
+    storage: {
+      strategy: 's3',
+      config: {
+        bucket: 's3-bucket',
+        region: 'eu-central-1',
+        accessKeyId: process.env.ENV_VAR,
+        secretAccessKey: process.env.ENV_VAR,
+        signatureVersion: 'v4',
+        httpOptions: {
+          agent: proxy('http://proxy.some-host.com:3128/')
```
marcbachmann: I don't think there's any need for `proxy-agent`.
You can either use the environment variable `HTTP_PROXY` or 
```
        httpOptions: {
          proxy: 'http://proxy.some-host.com:3128'
        }
```
or just
```
        proxy: 'http://proxy.some-host.com:3128'
```

marcbachmann: I think this is what we're using at bluewin.

gabrielhase:
@peyerluk do you want to keep this PR? quite old...

marcbachmann:
We've already imprved that in https://github.com/livingdocsIO/livingdocs/pull/224

PR #178, opened 2018-05-17
docs: new SSO

mthpvg:
@marcbachmann can you review?


gabrielhase:
Can we keep both documents? The earlier document states how to do *any* SSO using custom code. This is still useful as the core only allows for some SSO providers to be configured.

mthpvg:
@mthpvg please resume that PR.

mthpvg:
@gabrielhase can you take a look. I reverted the changes I did on the earlier document. Now, there is only a new one fitting in the server-configuration.

PR #179, opened 2018-05-19
Add contentType option in filters

gabrielhase:


gabrielhase:
@peyerluk you forgot this one

PR #180, opened 2018-05-25
Documentation for release-2018-05

DaRaFF:


reference-docs/server-configuration/admin-commands.md
```diff
@@ -37,3 +38,35 @@ grunt search-index:<index-name>:reset
...
+[tom, jerry, tom, tom, jerry, jerry] --> [tom, jerry, tom, jerry]
+[tom, jerry, donald, dagobert] --> [tom, jerry, donald, dagobert]
+```
+
+After that cleanup the system is still fully functional but needs a lot less space.
+
+### Cleanup Command Examples
+```bash
+# Show help
+node ./node_modules/@livingdocs/server/bin/tasks/cleanup
```
marcbachmann: Please do not mention internal structures like that.
Customers should use `livingdocs-server cleanup` as command.

DaRaFF: It's not available in the commands with livingdocs-server

marcbachmann: Why not? If you want to expose it, please add it.

PR #190, opened 2018-07-27
Documentation for release-2018-08

DaRaFF:


reference-docs/server-configuration/config.md
```diff
@@ -18,7 +18,8 @@ server: {
 ```js
 editor: {
   // configure the Editor Host. This is used for password reset & login urls.
-  public_host: 'http://localhost:9000'
+  public_host: 'http://localhost:9000',
+  auto_save_interval: 10 // This defines how often the editor auto saves in seconds
```
DaRaFF: @ReicoC `release-2018-08` is an integration branch, and a change also needs a PR. But for this time it's ok :-)

ReicoC: oh sorry, will do so next time

marcbachmann: @DaRaFF Github worked a bit on the branch checks. Maybe they support a regex now which we could use to protect those branches.

marcbachmann: Only one pattern is supported, but we can match against `release*` with a second check
image

PR #195, opened 2018-09-23
Document usage of Upstream factories in Downstream project

munen:


reference-docs/writing-tests/tests-with-factories.md
```diff
@@ -0,0 +1,73 @@
...
+To give you a feel on how to use the factories, here is a working
+example which tests the expected side-effects within an LI Server when
+creating a new `DocumentEntity`. The test itself is rather mundane, it
+checks that an entry with the `documents` table has been created while
+there is no entry in the `document_publication_events` table, because
+the document hasn't been published, yet.
+
+```js
+const DocumentEntity = require('@livingdocs/server/app/features/documents/entities/document_entity')
+const factories = require('@livingdocs/server/test/support/factories')
```
peyerluk: I like showing how to use the factories. But as far as I see the test tests upstream functionality and has to tap into the upstream internals via the `DocumentEntity`. Ideally the downstreams work through APIs and exposed features whenever possible. Deep requires into `@livingdocs/server/app/features/...` are more of an anitpattern where it can be avoided.

Do you have an idea for another example where we test some downstream functionality?

munen: For clarification: It would be OK to deep require `@livingdocs/server/test/support/factories` for an example, though?

peyerluk: We should add everything that is required from a downstream to our exports file: https://github.com/livingdocsIO/livingdocs-server/blame/master/exports/index.js#L4

Because even for an example a deadrequire is no fun.

munen: Then I suggest to close this PR without merging.

PR #204, opened 2019-01-09
feat(windows): add documentation for developing livingdocs in windows 10

meinradandermatt:
We added the documentation for running livingdocs locally on windows 10.


mokimo:
Follow-up suggestions from slack: add running `docker-compose up -d` under windows and not the bash

PR #208, opened 2019-01-17
Add image source policy docs

peyerluk:


reference-docs/editor-configuration/image-source-policy.md
```diff
@@ -0,0 +1,38 @@
...
+    }]
+  }
+}
+```
+
+images.sourcePolicy is a new config option which is optional. By default upload and hugo are enabled and url is disabled.
+
+Current default policy:
+
+const defaultPolicies = {
```
DaRaFF: You should add backtickts to the config.

PR #212, opened 2019-02-12
Add multi-select documentation

gabrielhase:


gabrielhase:
@peyerluk this is a small one ;)

PR #214, opened 2019-02-12
Add data-migration examples to the enterprise documentation 

mokimo:
## Overview:
Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/2418

## Description:
We want to remove the data migration examples from the boilerplate and instead have the examples within our enterprise documentation

## Relation: 
This should be merged first: https://github.com/livingdocsIO/livingdocs-server-boilerplate/pull/132

### DIFF
--- a/concepts/document-migrations/examples/remove_directive.md
+++ b/concepts/document-migrations/examples/remove_directive.md
@@ -1,0 +1,80 @@
+## Removing a directive
+
+Removing a directive in a design is a breaking change.
+To denote it in the design you should use a major semver version (http://semver.org/). In any case you need to write a data migration in order to be able to open old documents with the new design. The below code is a sample migration for a design change that removes an editable directive called "author" from a component. 
+
+_You will first need to configure a design for the migrations to work._
+
+## Further reference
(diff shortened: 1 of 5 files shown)


concepts/document-migrations/examples/remove_directive.md
```diff
@@ -0,0 +1,80 @@
...
+//  serializedLivingdoc: will overwrite the input
+//  metadata: will overwrite the input
+//
+// The grunt task 'data-migrations' will call this migration method with every
+// document of the project you specify in the task.
+exports.migrate = function ({serializedLivingdoc, metadata} = {}, callback) {
+  // You can work directly on the JSON or create a livingdoc instance from
+  // it and use the livingdoc API. We do the latter in this example.
+
+  const livingdoc = liSDK.document.create({
```
marcbachmann: Do we want to make an example that uses the framework?
Digging into the framework api is sometimes more complicated for customers than just explaining the json data structure. We can't change the api of the framework anyways without announcing a breaking change, therefore I'd introduce simpler operations in the future. Maybe we can make a list of the most common operations we'll need to support.

/cc @peyerluk

peyerluk: > Do we want to make an example that uses the framework?

You mean instead of using the SDK?

PR #221, opened 2019-02-14
Reorganisation of the navigation

peyerluk:
## Description

Reorganisation of the main Navigation. The main focus was on the reference documentation and to indicate the context of the different docs better. Also I tried to separate Concepts/Guides from Howtos. And a few small things like Title Casing all the links and renaming some entries.

This PR is best to review locally by running `gitbook serve` and checking out what you like, don't like in the navigation.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -2,9 +2,16 @@
 
 * [Getting started](./getting_started.md)
 
-* [Setting up Livingdocs locally](walkthroughs/getting-started-with-local-development.md)
+
+## Concepts
 
 * [Glossary](./DICTIONARY.md)
+* [Livingdocs Design and Components](reference-docs/common-designs/create_designs.md)
+* [Metadata Plugins](walkthroughs/metadata/metadata-examples.md)
+* [Image Services](concepts/images/image-services.md)
+* [Includes](reference-docs/doc-includes/intro.md)
+* [Document Copy](concepts/copying-documents/document_copy_feature.md)
+
 
 ## Videos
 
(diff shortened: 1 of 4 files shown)


SUMMARY.md
```diff
@@ -2,72 +2,81 @@
 
 * [Getting started](./getting_started.md)
 
-* [Setting up Livingdocs locally](walkthroughs/getting-started-with-local-development.md)
+
+## Concepts
```
gabrielhase: You had that before and I was against having a "concepts" section :) It's mainly about the name though, the section as such is fine. Looking at the content it's almost 100% the same as the Evaluation Guide from @meinradandermatt 
So my proposal would be:
- call this section "Evaluation Guide" or "Evaluation Intro"
- make the bullet point names actions, e.g. `Livingdocs Design and Components` -> `Learn how to create your design`
- Remove `Document Copy` from here and put this into the Editor reference (doesn't fit here IMO)
- Take in: Setting up locally (under core atm) and `Import your old documents` (don't know if we already have something for this otherwise we should write it)

SUMMARY.md
```diff
@@ -2,72 +2,81 @@
...
+  * [Image Cropper Config](reference-docs/editor-configuration/image-cropping.md)
+  * [Image Source Policy Config](reference-docs/editor-configuration/image-source-policy.md)
+  * [Metadata Configuration](reference-docs/editor-configuration/metadata.md)
   * [Login](reference-docs/editor-configuration/login.md)
-* [Events](reference-docs/server-extensions/events.md)
-* [Image cropper config](reference-docs/editor-configuration/image-cropping.md)
-* [Image source policy config](reference-docs/editor-configuration/image-source-policy.md)
-* [Image services](concepts/images/image-services.md)
-* [Import API](reference-docs/server-import-api/import_api.md)
+* Livingdocs Design
```
gabrielhase: @meinradandermatt you are looking into the design workflow right? Our design documentation is by far the sparsest :) maybe you can also think of doing something here

SUMMARY.md
```diff
@@ -2,72 +2,81 @@
...
+  * [Design Config](reference-docs/common-designs/design_config.md)
+  * [Component Config](reference-docs/common-designs/component_config.md)
+* Livingdocs Content Model
```
gabrielhase: I would rather call this `Livingdocs Content API` or `Content Object API`

SUMMARY.md
```diff
@@ -77,10 +86,10 @@
 
   * [Hardware Requirements](setup-and-deployment/hardware-requirements.md)
   * [Architecture](setup-and-deployment/high-availability/README.md)
-    * [high-availability](setup-and-deployment/high-availability/high-availability-setup.md)
+    * [High-Availability](setup-and-deployment/high-availability/high-availability-setup.md)
```
gabrielhase: @DaRaFF Is it fair to say that this documentation is outdated? Or can we still live with it?

SUMMARY.md
```diff
@@ -89,12 +98,10 @@
...
 
-## Administration
 
-* [Create Users](walkthroughs/create-users.md)
-* [Assign access rights](administration/access_rights.md)
-* [Logs](reference-docs/server-configuration/logging.md)
-* [Server Admin Commands](reference-docs/server-configuration/admin-commands.md)
+## Core Development
+
+* [Setup for Core Development](walkthroughs/getting-started-with-local-development.md)
```
gabrielhase: Setting up Livingdocs locally needs to go to the very top. It's one (if not the) first thing that customers do.

SUMMARY.md
DaRaFF: It would not say it's outdated, but it's very very very high level. This means the reader does not profit much from this perspective.

SUMMARY.md
meinradandermatt: I think my evaluation guide could be the start point for this section. We can link then from there to all the articles in this section.

SUMMARY.md
meinradandermatt: Yes, I think we should add the documentation about how to develop a design. We would like to change that so after the change we can add the documentation.

SUMMARY.md
peyerluk: I like the idea of having an 'Evaluation Guide' Section. I will look at the Evaluation Guide PR together with @meinradandermatt and we will merge these two PRs at the same time.

'Setting up locally' will in some form be part of the evaluation guide. But we'll have to rephrase a few things so they make sense in the boilerplate setup. The current wording is aimed at core devs.

'Document Copy' doesn't fit the reference section but maybe it fits in the howto section or we can split it up.

peyerluk:
I tried to address feedback if there are further improvements let's tackle them in new PRs.

PR #224, opened 2019-02-21
Add docs for configuring feature storage

peyerluk:
## Description

Add docs for the `storage` config option that is used in various features in the server config.

### DIFF
--- a/reference-docs/server-configuration/config-storage-options.md
+++ b/reference-docs/server-configuration/config-storage-options.md
@@ -1,0 +1,91 @@
+# Storage Config Options
+
+#### Description
+
+Various features have to store files in a blobStore. For this they offer a `storage`
+config option that can be configured to store data in memory, on the file system or in S3.
+
+Underlying this we use the abstractBlobStore API: https://github.com/maxogden/abstract-blob-store
+
+
+#### S3
+
+Standard S3 Config
+```js
+storage: {
+  strategy: 's3',
+  config: {
+    bucket: 'livingdocs-images-dev',
+    params: {ACL: 'public-read'},
+    region: 'eu-west-1',
+    accessKeyId: 'key',
+    secretAccessKey: 'secret'
+  }
+}
+```
+Note: `params: {ACL: 'public-read'}` is set by default if you omit `params`.
+
+For more information about the config object see the [S3 docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#constructor_details).
+
+#### S3 With A Proxy
+
+The standard S3 config will respect the `HTTPS_PROXY` environment variable.
+(e.g. `HTTPS_PROXY=http://localhost:3128`)
+
+Or you can specify a proxy explicitly as seen in the examples below.
+
+
+S3 with a proxy set via `httpOptions`:
+```js
+storage: {
+  strategy: 's3',
+  config: {
+    bucket: 'livingdocs-images-dev',
+    region: 'eu-west-1',
+    accessKeyId: 'key',
+    secretAccessKey: 'secret',
+    httpOptions: {
+      proxy: 'http://localhost:3128'
+    }
+  }
+}
+```
+
+S3 with a `proxy-agent` proxy set via `httpOptions`:
+```js
+storage: {
+  strategy: 's3',
+  config: {
+    bucket: 'livingdocs-images-dev',
+    region: 'eu-west-1',
+    accessKeyId: 'key',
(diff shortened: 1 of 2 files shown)


reference-docs/server-configuration/config-storage-options.md
```diff
@@ -0,0 +1,95 @@
...
+S3 with a proxy:
+```js
+storage: {
+  strategy: 's3',
+  config: {
+    bucket: 'livingdocs-images-dev',
+    region: 'eu-west-1',
+    accessKeyId: 'key',
+    secretAccessKey: 'secret',
+    proxy: 'http://localhost:3128'
```
marcbachmann: Should we deprecate that property and always suggest `httpOptions.proxy`?

reference-docs/server-configuration/config-storage-options.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Local File System
+
+Example setup to write to a temporary folder:
+```js
+storage: {
+  strategy: 'fs',
+  config: {
+    path: require('os').tmpdir(),
+    bucket: 'test-bucket'
```
marcbachmann: I don't think that `bucket` is needed or even supported here.

reference-docs/server-configuration/config-storage-options.md
peyerluk: I don't know the difference really... 
Should we remove this example from the docs or just add a comment to it?

marcbachmann: I'd remove the example. We currently just write `proxy` onto `httpOptions.proxy`.

peyerluk:
@marcbachmann Updated

reference-docs/server-configuration/config-storage-options.md
```diff
@@ -0,0 +1,87 @@
...
+    accessKeyId: 'key',
+    secretAccessKey: 'secret'
+  }
+}
+```
+
+#### S3 With A Proxy
+
+The standard S3 config will respect the `HTTPS_PROXY` environment variable.
+(e.g. `HTTPS_PROXY=http://localhost:3128`)
```
marcbachmann: Currently we don't have a good solution to override that HTTPS_PROXY variable.
This is currently a pain in case customers don't want to use a proxy but the variable is declared.

reference-docs/server-configuration/config-storage-options.md
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+}
+```
+
+S3 with a `proxy-agent` proxy set via `httpOptions`:
+```js
+storage: {
+  strategy: 's3',
+  config: {
```
marcbachmann: Should we link to the aws-sdk s3 documentation for all the options?

peyerluk: Yes, that's a good idea. Do you have the link at hand?

marcbachmann:
Currently we always set the `ACL` to `public-read`. We should document that or change the defaults. https://github.com/livingdocsIO/livingdocs-server/blob/master/lib/blob-store/strategies/s3.js#L4

I think we don't rely on that anymore as `imgix` directly reads from the s3 buckets with credentials.

reference-docs/server-configuration/config-storage-options.md
marcbachmann: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#constructor_details

peyerluk:
I added the `ACL` param to the docs.

reference-docs/server-configuration/config-storage-options.md
```diff
@@ -0,0 +1,91 @@
...
+    path: require('os').tmpdir()
+  }
+}
+```
+
+
+#### Memory
+
+For testing purposes you can gain performance with a memory storage:
```
marcbachmann: Not only performance, but mainly isolation and a temporary storage, which is the main reason we use that.

reference-docs/server-configuration/config-storage-options.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+#### Local File System
+
+Example setup to write to a temporary folder:
+```js
+storage: {
+  strategy: 'fs',
+  config: {
+    path: require('os').tmpdir()
```
marcbachmann: btw. `require('os').tmpdir()` returns a string to a temporary directory which gets removed upon server restart or docker container restart.

PR #226, opened 2019-02-28
feat(add-admin-tasks): add task search-index and create Admin User task

meinradandermatt:
# Goal
Add missing tasks to the documentation. The tasks are described also in the readme of the boilerplate.

### DIFF
--- a/reference-docs/server-configuration/admin-commands.md
+++ b/reference-docs/server-configuration/admin-commands.md
@@ -20,9 +20,9 @@ There are a lot of commands which can be executed on the server.
 
 `grunt project-create` creates a new project.
 
-
 ## <a name="search-index">Maintain Search Index</a>
 
+`livingdocs-server search-index` is a tool to recreate and reindex a search index.
 `grunt search-index` is a tool to create, index, update and delete a search index.
 
 ### Examples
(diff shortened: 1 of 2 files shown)


walkthroughs/create-users.md
```diff
@@ -5,6 +5,26 @@ To set up the **initial user**, run the following command in the project folder.
 ```sh
 grunt user-create-admin
 ```
+or
```
DaRaFF: maybe it would help, if you describe here what an admin user is and what he can and can't do (e.g. login into the editor and work in a specific project)

reference-docs/server-configuration/admin-commands.md
```diff
@@ -20,9 +20,9 @@ There are a lot of commands which can be executed on the server.
 
 `grunt project-create` creates a new project.
 
-
 ## <a name="search-index">Maintain Search Index</a>
 
+`livingdocs-server search-index` is a tool to recreate and reindex a search index.
 `grunt search-index` is a tool to create, index, update and delete a search index.
```
marcbachmann: We should port most of the features into the new task instead of listing the grunt setup, which requires customers to configure grunt in their setup.

gabrielhase:
Do we still want to merge this?

PR #243, opened 2019-05-16
Add editor dashboard config documentation

peyerluk:


reference-docs/editor-configuration/menu-and-dashboards.md
```diff
@@ -301,3 +316,109 @@ liEditor.searchFilters.register('project', ['session', (session) => ({
   }
 })])
 ```
+
+
+Relations:
```
DaRaFF: I don't think we should link to issues and define tbd in our documentation.

reference-docs/editor-configuration/menu-and-dashboards.md
```diff
@@ -301,3 +316,109 @@ liEditor.searchFilters.register('project', ['session', (session) => ({
...
+  - Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/2628
+  - Documentation: tbd
+
+
+## Custom Dashboards
+
+Custom Dashboards are configurable versions of the current Articles / Pages / Date-Records Screens. You can as many custom dashboards as you want in the Editor configuration.
```
DaRaFF: You can as many... -> You can add as many...

reference-docs/editor-configuration/menu-and-dashboards.md
```diff
@@ -301,3 +316,109 @@ liEditor.searchFilters.register('project', ['session', (session) => ({
...
+  // CSS class set as a wrapper around the result list
+  cssWrapper: 'li-result-columns'
+}]
+```
+
+Kanban Boards are very similar to dashboards, except they do have multiple result columns. Each result column will show a list of documents the same as a single dashboard does. The documents cannot be manually sorted or moved between columns, instead each column typically has its own filter settings.
+
+For example a task board will show all tasks in the `requested` state in one column and tasks with the state `inProgress` and `done` in the other columns. In order to move a card into another column you simply have to open the document and move the task into another state.
+
+Simple Task KanbanBaord Config:
```
DaRaFF: Board

PR #248, opened 2019-06-20
Use new config format for metadata examples

peyerluk:
## Changelog

Update metadata examples to use the new config format (channelConfig v2)

### DIFF
--- a/reference-docs/editor-configuration/metadata.md
+++ b/reference-docs/editor-configuration/metadata.md
@@ -30,16 +30,21 @@ The following subchapters discuss them and show the available options.
 
 ### Text Input
 
+ContentType metadata config:
 ```js
-name: 'nameOfYourMetadataTextField',
-form: 'li-meta-text-form',
-config: {
-  service: 'nameOfYourServicePlugin', // mandatory
-  label: 'foo', // optional, takes camelized name otherwise
-  placeholder: 'bar', // optional, takes camelized name otherwise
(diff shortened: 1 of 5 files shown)


peyerluk:
@mokimo Do you want to have a look at the second commit?

reference-docs/editor-configuration/metadata.md
```diff
@@ -30,16 +30,21 @@ The following subchapters discuss them and show the available options.
...
+metadata: [{
+  ...,
+  ui: {
+    component: 'liMetaTextForm',
+    config: {
+      label: 'foo', // optional, takes camelized name otherwise
+      placeholder: 'bar', // optional, takes camelized name otherwise
+      readOnly: false, // optional, false by default
+      maxLength: 200, // optional, integer, not used by default
+      service: 'nameOfYourServicePlugin' // optional
```
mokimo: Can you check if `service` is within ui or ui.config? As far as I've found out it's outside of it, but seen mixxed results:
https://github.com/livingdocsIO/livingdocs-server/blob/master/app/features/channel-configs/tests/fixtures/channel_config_v2.js#L56

mokimo: https://github.com/livingdocsIO/livingdocs-editor/blob/master/app%2Ffeatures%2Fmetadata%2Fmetadata_forms%2Fmetadata_form.js#L82

`ui.config.service` should be `ui.service`

PR #252, opened 2019-07-09
update differ config

meinradandermatt:
# Description
We doesn't support `diffViewEnabled` anymore. So we had to remove it from the documentation.

### DIFF
--- a/reference-docs/editor-configuration/editing-features.md
+++ b/reference-docs/editor-configuration/editing-features.md
@@ -224,7 +224,6 @@ Both options are recommended.
 ```
 app: {
   history: {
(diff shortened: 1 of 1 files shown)


reference-docs/editor-configuration/editing-features.md
```diff
@@ -224,7 +224,6 @@ Both options are recommended.
 ```
 app: {
   history: {
-    diffViewEnabled: true,
```
DaRaFF: Is this somewhere mentioned in the corresponding PR that this property was removed? This needs to be mentioned in the release notes, so the customers have the chance to remove this property in their config.

PR #255, opened 2019-08-14
Add example for rawDocument in importApi

peyerluk:


marcbachmann:
I just saw that there's no mention of `contentType` in the docs.... only `documentType`.
Isn't the `contentType` is mandatory? 😁🤔 
The `documentType` is basically redundant as we could figure that out based on the `contentType`.

peyerluk:
> Isn't the contentType is mandatory

It should be... you're right.

PR #257, opened 2019-08-20
Documentation: Option to Opt In for Auto Update of `doc-html` Input

disufi:
Implementation: https://github.com/livingdocsIO/livingdocs-editor/pull/2917


disufi:
This will be the default, config is gone now.

PR #260, opened 2019-09-16
doc: Using a local design

munen:
Taken straight from my notes that I created to be able to develop/run a design locally.

If you're inclined to add this to the documentation, I can also polish the wording. I didn't want to spend a lot of time on this (without being asked), I just wanted to put it here, because I did actually get asked how to do this^^


munen:
@gabrielhase Thanks for letting me know. I'll close the PR, then!

PR #263, opened 2019-09-27
Document categories

gabrielhase:


reference-docs/server-configuration/config.md
```diff
@@ -406,17 +406,54 @@ render_pipeline: {
...
+    enabled: true,
+    debug_routes: false,
     // number of publication events to process in each batch
     batch_size: 50000, // (default: 50000)
     // routes cache update interval
     watch_interval: 1000 // (default: 1000)
   },
-  db: 'leveldown' // (default: leveldown) to use leveldb, npm i leveldown and set 'leveldown' here
-  // db_options: # (optional) cf. https://github.com/Level/levelup#options
+  redis: {
```
DaRaFF: is that correct? shouldn't it be `db: 'redis'`?

Anyway, it would definetively help when the supported options are documented in a comment.

gabrielhase: I hope it is, this is from our production config for the service :)
redis should be the only options for productive use, I also write that later in the plain text. We don't want to support other options

marcbachmann: Yes, redis as object is correct. We have a weird combination of `routing.redis` and `kv.redis` configs. We should focus more on easier and more straight forward configurations in the future.

PR #278, opened 2020-02-20
extend service config

meinradandermatt:
# Description
The service documentation is extended and parts like the possible design settings, component options and channel config options are added. 

Some parts are still linked to the enterprise documentation which is not perfect. For example the metadata plugins should be a own page for the service.


service/project_setup.md
```diff
@@ -1,123 +1,39 @@
-# Get Started with the Service
-With the following steps a project can be setup and used with a own design.
+# Get Started with Livingdocs as a Service
+Livingdocs offers a headless CMS as a service.
+
+This short guide explains how to set up and use the service with your own design.
```
mokimo: Should we make the differentiation between CLI-setup and just clicking through the UI? The easiest use-cases should later be possible without diving too deep the _whole_ setup using the CLI imo

service/project_setup.md
```diff
@@ -1,123 +1,39 @@
...
 - Invite Users to your project
 
-# Design setup
-Use the https://github.com/livingdocsIO/magazine-example Design example to see a design example. This is not an embedded Design so it can't be used the same way as an embedded design.
-
-The config options for a design: [design configs](../reference-docs/common-designs/component_config.md)
-
-### project config download and changes
+## project config download and changes
 With the following steps an embedded Design can be created for the new project.
```
mokimo:
```suggestion
With the following steps an embedded Design can be created for the new project using the livingdocs-CLI.
```

service/project_setup.md
```diff
@@ -1,123 +1,39 @@
...
 
-## Design restrictions
-At the moment following changes on the design are not supported.
-- rename a directive in a component
-- rename a component
-- add a new directive to a component
 
+## Restrictions
+- content-types can't be removed
+- metadata-fields can not be removed
```
mokimo: - content-types can't be removed/renamed
- metadata-fields can't be removed/renamed
- components can't be removed/renamed 
  (You can remove a component from a content-type, but not from the design itself)

Should you run into an error and you can't open the documents anymore, please go the `project-settings/design migration` tab in the editor and manually downgrade the design version of your documents.

service/project_config.md
```diff
@@ -0,0 +1,13 @@
+# project config
+The project can have specific configurations and content-types. 
+
+## settings
+In the `settings.js` file of a project the configuration can be set for the channel (project). On the following link the possible configurations are shown: 
+[channel configs](../reference-docs/server-configuration/channel-config.md)
+
+## content-types
+A project can have multiple content-types. Usually individual content-type is created if the default content looks very different and has a different set of components which can be used.
```
mokimo: A typical example use case for would be having a `regular`-article content type which would be for writing stories.
Another content-type could be a `gallery`-article, which would display a lot of images or galleries instead of text.

Each content-type can have it's own set of default components, available components, metadata and settings.

__

@meinradandermatt do you think this makes sense? Our service customers have no idea about the concept of a content-type, content-types can be pretty confusing coming in touch with livingdocs for their first time, so examples may help

service/design_settings_config.md
```diff
@@ -0,0 +1,107 @@
...
+```
+  properties: ['css-background-color', 'css-class', 'css-class-selection']
+```
+## assets
+You can add custom css and js files to your design. These files will be loaded in the editor to show an article.
+
+```
+assets: {
+    css: [
+      'url to css'
```
mokimo: 'http://example.com/foo.css'

service/design_settings_config.md
```diff
@@ -0,0 +1,107 @@
...
+    ], 
+    js: [
+        'url to js'
```
mokimo: 'http://example.com/foo.js'

PR #282, opened 2020-02-29
Add mapping example of v1 to v2 design config

DaRaFF:


DaRaFF:
@peyerluk I need your help. I wrote down the mapping of a v1 to a v2 design config and I am not sure, if I'm correct. Can we sit together and look into it?

PR #286, opened 2020-03-16
New Login Device Detection

disufi:
Adds documentation.

### DIFF
--- a/reference-docs/server-configuration/config.md
+++ b/reference-docs/server-configuration/config.md
@@ -192,6 +192,46 @@ emails: {
 }
 ```
 
+##### New Login Device Detection
+
+In case you want to activate abuse detection through login device comparison, please configure:
+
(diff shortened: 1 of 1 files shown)


reference-docs/server-configuration/config.md
```diff
@@ -192,6 +192,46 @@ emails: {
...
 
+##### New Login Device Detection
+
+In case you want to activate abuse detection through login device comparison, please configure:
+
+```js
+auth: {
+  connections: {
+    local: {
+      sendNotificationIfNewDevice: true
```
DaRaFF: do you validate in the code if this setting is activated that the email template is registered?

disufi: No, but it doesn't crash. The server logs out that the template was not registered through the email api.

PR #287, opened 2020-03-25
feat(integrations): add documentation for the new integrations feature

benib:
Relations:
  - Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/3359
  - Related PR:
    - https://github.com/livingdocsIO/livingdocs-server/pull/2849
    - https://github.com/livingdocsIO/livingdocs-editor/pull/3333

### DIFF
--- a/reference-docs/server-configuration/config.md
+++ b/reference-docs/server-configuration/config.md
@@ -725,7 +725,25 @@ The credentials object is just the google service account json. We advise you to
 
 ## Integrations
 
+Added in: `release-2020-04`
+
+There is a general `integrations` feature for small integrations that can be configured in the `channelConfig` through editor UI. They need to be allowed in the server config with the following config.
+
+#### Comyan
+
+To allow the `comyan` integration, add this to your config:
+
+```js
+integrations: {
+  comyan: {
+    allowed: true
+  }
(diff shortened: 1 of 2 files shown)


reference-docs/server-configuration/config.md
```diff
@@ -725,6 +725,24 @@ The credentials object is just the google service account json. We advise you to
...
+
+```js
+integrations: {
+  comyan: {
+    allowed: true
+  }
+}
+```
+
+## Big Integrations
```
DaRaFF: Big integrations is a strange word :-). Maybe you could call this 2 topics "Standard Integrations" and "Other Integrations"? In the long run we want to get rid of the "Other Integrations" anyway.

benib: In the design for the integration feature we talked about small things where it would be overkill to create a full blown feature like hugo.
I agree that the wording is weird.
To me standard and other is a little off as well, since it is not clear if we really want to move big things like hugo over.

Maybe we just keep Hugo under integrations without a separate subtitle?

DaRaFF: Yep, sounds good.

PR #289, opened 2020-03-27
fix(cli): add docs for config-sync workflow

peyerluk:
## Changelog

- Add docs how to set up a `.livingdocs-cli` file
- Describe config sync process with the CLI
- Add those to the main navigation

### DIFF
--- a/livingdocs-cli/cli-dotfile.md
+++ b/livingdocs-cli/cli-dotfile.md
@@ -1,0 +1,110 @@
+# .livingdocs-cli
+
+## Set up a cli dotfile
+
+In case you are working with multiple projects and/or environments it becomes
+cumbersome to set tokens, hosts and other params.
+
+To make this easier you can set up a dotfile that allows you to pass
+`env` and `project` to all commands where this makes sense instead of
+using e.g. the `LI_TOKEN` environment variable.
+
+
+#### Multiple environments
+
+`.livingdocs-cli` file in the root of your working directory:
+```json
+{
+  "environments": {
+    "local": {
+      "host": "https://localhost:9000",
+      "token": "local-token",
+      "distFolder": "./config-sync/local"
+    },
+    "development": {
+      "host": "https://server-development.example.dev",
+      "token": "dev-token",
+      "distFolder": "./config-sync/development"
+    },
+    "staging": {
+      "host": "https://server-staging.example.dev",
+      "token": "staging-token",
+      "distFolder": "./config-sync/staging"
+    },
+    "production": {
+      "host": "https://server-production.example.dev",
+      "token": "production-Token",
+      "distFolder": "./config-sync/production"
+    }
+  },
+  "alias": {
+    "dev": "development"
+  }
+}
+```
+
+How to use these configs in the terminal:
+```sh
+# with `env` param
+npx livingdocs-cli project-config:download --env development
+
+# in shorthand form
+npx livingdocs-cli project-config:download -e development
+
+# in shorthand form and also using an alias
+npx livingdocs-cli project-config:download -e dev
+```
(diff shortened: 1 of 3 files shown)


livingdocs-cli/cli-dotfile.md
```diff
@@ -0,0 +1,110 @@
+# .livingdocs-cli
+
+## Set up a cli dotfile
+
+In case you are working with multiple projects and/or environments it becomes
+cumbersome to set tokens, hosts and other params.
+
+To make this easier you can set up a dotfile that allows you to pass
```
DaRaFF: To make the sync process easier...

livingdocs-cli/cli-dotfile.md
```diff
@@ -0,0 +1,110 @@
...
+      "distFolder": "./config-sync/development"
+    },
+    "staging": {
+      "host": "https://server-staging.example.dev",
+      "token": "staging-token",
+      "distFolder": "./config-sync/staging"
+    },
+    "production": {
+      "host": "https://server-production.example.dev",
+      "token": "production-Token",
```
DaRaFF: I think this is problematic because most customers don't want to check in a prod token. For the future, it would be cool if one could define an environment variable in `.bashrc` e.g. `MY_SUPER_ENV` and connect this variable here e.g. `"token": "$MY_SUPER_ENV"`.

livingdocs-cli/sync-configs.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### Full multi-project enterprise workflow example
+
+This workflow is for enterprise customers who also have a local setup
+of livingdocs on their machine.
+
+
+1. Download all configs:
+```sh
+npx livingdocs-cli project-config:download --format json -p projectA --env dev &&
```
DaRaFF: For the future: `npx livingdocs-cli project-config:download --format json --all` would be cool :-)

livingdocs-cli/sync-configs.md
```diff
@@ -0,0 +1,96 @@
...
+
+5. Make changes to config files
+
+6. Publish changed configs (also just one example)
+```sh
+# plan update and verify only the expected properties are updated
+npx livingdocs-cli project-config:plan -p projectA --env development
+
+# if everything looks ok, do the update
+npx livingdocs-cli project-config:publish -p projectA --env development
```
DaRaFF: For the future: this could be a tricky part for customers regarding timing. When do we execute this command? Before/after the deployment of the new version on a specific environment? Can we automate that step (usually they want to deploy a new version and not do an extra step to publish the configs)

livingdocs-cli/sync-configs.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+7. Discard `sync-projectA/local` and `sync-projectB/local`
```
DaRaFF: Why discarding the local config? I guess I updated the local config and the updated the other environments. Why I want to loose everything on local?

livingdocs-cli/cli-dotfile.md
peyerluk: You can always leave this away and supply a toke via the --token argument. I'll extend the description a bit with the recommendation to only supply a read token here for production.

livingdocs-cli/sync-configs.md
peyerluk: or `npx livingdocs-cli project-config:download --format json --env *`

livingdocs-cli/sync-configs.md
peyerluk: I just thought you dont really need to track the local config. But there is no harm in keeping it either :)

PR #303, opened 2020-06-12
remove and restructure (but mostly remove)

gabrielhase:
This PR mainly does 2 things:
- remove outdated stuff (lots of it)
- restructure the docs to a much simpler structure. best to read it here: https://github.com/livingdocsIO/livingdocs/pull/303/commits/8d3e2c3912f5ebb3543a05d8773bbe68d0fea58b

The link above shows the new contribution guidelines. We should agree on those, then we should enforce them going from here.


DaRaFF:
As already said.... after merging this PR you should test if the links still work. It's not possible to check that via PR review.

PR #314, opened 2020-07-29
doc(includes): improve includes documentation

benib:
1. this changes the order of include ui customizations as follows to better reflect our preferences:
  - vue-component (new)
  - iframe-modal (added doc about postMessage interface for passing `params`)
  - angular-component (untouched)
  - angular-modal (untouched)

2. the vue component type `includeParamsSidebarForm` is added to the vue registry docs

### DIFF
--- a/reference-docs/includes/server_customization.md
+++ b/reference-docs/includes/server_customization.md
@@ -185,50 +185,36 @@ module.exports = {
 The `uiComponents` array allows you to define a list of ui elements that are rendered in the sidebar upon selection of the `doc-include` (top to bottom). As you register your own component or helper provided by livingdocs, you have more choices but the development time increases as well. 
 
 You can choose between 3 types of custom UI components.
-1. `angular-component`, sidebar user interface
 
-2. `angular-modal`, user interface in a modal dialog (if more space is required), button in the sidebar to open the modal
+1. `vue-component`, sidebar user interface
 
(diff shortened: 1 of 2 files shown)


reference-docs/includes/server_customization.md
```diff
@@ -185,68 +185,105 @@ module.exports = {
...
 
-2. `angular-modal`, user interface in a modal dialog (if more space is required), button in the sidebar to open the modal
+1. `vue-component`, sidebar user interface
 
-3. `iframe-modal`, as above user interface in a modal dialog but loaded as an iframe (e.g. if you want to implement your UI outside of Livingdocs)
+2. `iframe-modal`, as above user interface in a modal dialog but loaded as an iframe (e.g. if you want to implement your UI outside of Livingdocs)
 
-The first two require an angular plugin to be present in the Livingdocs editor.
-The markup for `angular-component` looks as follows:
+3. `angular-component`, sidebar user interface
```
mokimo, line 193: Do we want to add a note that they are deprecated or in LTS since `release-2020-07` ?
They are soft-deprecated now that we have a vue alternative IMO

benib: I would not deprecate it just now. If we start to have a clear path for deprecation I am all for it, but we should have a rough timeline and logs from within the code and also a documented path for migration. Currently we can't cover all the usecases of these include components in Vue just yet.

PR #315, opened 2020-08-03
add text-formatting documentation

meinradandermatt:
Add the text-formatting configs to the documentation.


reference-docs/project-config/content_types.md
```diff
@@ -327,4 +327,55 @@ metadata: [{
 }]
 ```
 
-With this in place you can set the channel configuration for your push notification topics \(see example config above\) and the firebase configuration in the [server config](../server-configuration/config.md#push-notifications).
\ No newline at end of file
+With this in place you can set the channel configuration for your push notification topics \(see example config above\) and the firebase configuration in the [server config](../server-configuration/config.md#push-notifications).
+
+## Text Formatting
+The text formatting toolbar (the toolbar which is shown after selection of a text) can be customized. When this configuration is added the default configuration and the configuration from the project will be overwritten.
```
mokimo, line 333: A little add-on would be adding a gif here. Definitely not necessary if you're done and want to merge it.

But if you have a few extra minutes showing the custom settings in the service project in action that would be cool!

gabrielhase:
@meinradandermatt for the future please merge all service relevant documentation (developers.livingdocs.io) also to the branch `service` (https://github.com/livingdocsIO/livingdocs/tree/service). If you have this in mind from the beginning then you can adjust your commits in that way and use cherry picking.
Nothing to do for this one, I'll take it over.

PR #330, opened 2020-10-19
Document image service 2.0

peyerluk:


guides/image-services.md
```diff
@@ -2,85 +2,155 @@
 
 ## Overview
 
-Livingdocs allows users to upload pictures and then subsequently shows them. There are 3 parts here:
+For Livingdocs images are part of the document and Livingdocs supports mechanisms
```
gabrielhase, line 5: I wouldn't write "images are part of the document". That kind of undermines our media library. I would rather write: "Livingdocs documents are self-sufficient which means that they also contain all necessary image information. ..."

gabrielhase (approved):
I understand what's written but it seems not to be very stable IMO.
One can add its own image service via the framework but then you expect render strategies in the chosen configuration. There is no way to make sure that a custom image service actually applies to the chosen strategy. Or am I missing something?

guides/image-services.md
```diff
@@ -177,4 +234,36 @@ module.exports =
...
+Important: You will have to add image services to both your `server` and `editor`.
+
+An image service in livingdocs is basically just a url transformer that
+allows to proxy image requests through a 'real' image service like imgix and
+write params to the url dynamically to define e.g. the `width` or crop
+of an image.
+
+Example of registering an image service:
+```js
+const livingdocs = require('framework')
```
DaRaFF, line 251: const livingdocs = require('@livingdocs/framework')

DaRaFF: Generally, customers should not access to the framework directly. 
At the moment the only 2 solutions I know of are these:
- Add the framework to the downstream package.json
- Make a node hack a la https://github.com/nzzdev/livingdocs-api/blob/master/app/framework-extensions/framework_path.js

We want to prevent both of that for different reasons. We should write a wrapper function on the server and the editor, to allow downstreams to integrate a new image service.

With you suggested solution to create a separate npm package (which is indeed a good idea), we have the same issue as requiring the framework via package.json. You always have to keep the framework version up to date separately. In the past, we had a few issues when the framework versions did not match.

DaRaFF (approved):
👍  thx!

peyerluk:
@gabrielhase 

> One can add its own image service via the framework but then you expect render strategies in the chosen configuration. There is no way to make sure that a custom image service actually applies to the chosen strategy. Or am I missing something?

All render strategies work with all image services in this new setup. 

An image service does not need to know or do anything to support a specific render strategy except supporting the `width` parameter when creating urls.

(In the future there might be a few more like supporting to convert an animated gif to video for example)

guides/image-services.md
peyerluk: I reworded a bit.

guides/image-services.md
```diff
@@ -177,4 +234,36 @@ module.exports =
...
+livingdocs.imageServices.add({
+  name: 'myImageService',
+  supportsCrop: true,
+  supportsWidth: true,
+
+  // This is not a real-world example. It just rewrites
+  // the url to go through myproxy.com and embeds the width
+  // and image url in the path of the generated url.
+  getUrl: function (imageUrl, {crop, width}) {
+    return imageUrl || ''
```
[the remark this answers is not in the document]
peyerluk: That's the expected behavior right now. But you're right that could be improved.

guides/image-services.md
peyerluk: > With you suggested solution to create a separate npm package

The separate npm package would only contain the image service and not have a dependency to the framework.

I added an example for this to the docs as well.

DaRaFF (approved):
👍

PR #340, opened 2020-11-19
add airship documentation

meinradandermatt:
# Motivation

Add the documentation for the urban airship implementation

### DIFF
--- a/reference-docs/server-configuration/config.md
+++ b/reference-docs/server-configuration/config.md
@@ -555,9 +555,12 @@ For inspiration, you can also check out our [current default document search fun
 
 
 #### Push Notifications
+Configuration examples for the different notification services. 
 
+Firebase: 
 ```js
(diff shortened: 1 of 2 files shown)


reference-docs/server-configuration/config.md
```diff
@@ -582,6 +585,36 @@ push_notifications: {
 }
 ```
 
+Airship: 
+
+Added in: `release-2020-10`
+```js
+push_notifications: {
```
DaRaFF, line 592: We should always use camel case -> `pushNotifications`. The config can handle both, but we agreed on using camel case as standard.

DaRaFF (approved):

reference-docs/server-configuration/config.md
meinradandermatt: I change the firebase config too.

PR #343, opened 2020-11-25
doc(list include): improve list include example

benib:
This improves the list include example with code how to get documents from a list id.


gabrielhase (approved):
👍

PR #344, opened 2020-11-30
doc(includesApi): fix examples

benib:
## Motivation

We don't expose any `includesApi.registerServices` function to register an array of services but only `includesApi.registerService` to register a single include service.

This PR fixes the examples accordingly.


benib:
The examples are correct. In the server we expose `registerServices`

PR #347, opened 2020-12-16
chore(links): Fix broken links

ajwild:
I used the markdown-link-check tool mentioned in CONTRIBUTING.md to do most of the work.

I fixed every broken link except a Dokku file which no longer seems relevant. Hopefully I pointed everything to the correct location! Feel free to read through the sentences and see if the new links make sense, but a lot were obvious so it should be an improvement regardless of whether a couple are slightly wrong.


gabrielhase (approved):
Looks good 👍 
If you also want to fix it for developers.livingdocs.io you need to open a PR agains the branch `service`

PR #348, opened 2020-12-17
chore(links): Fix broken service links

ajwild:
More of the same link fixes as in PR #347, but for the service branch this time. I couldn't fix as many links as I managed in the other PR as there seems to be some missing documentation regarding Elasticsearch indexing, webhooks, and metadata.

### DIFF
--- a/getting_started.md
+++ b/getting_started.md
@@ -14,11 +14,11 @@ This short guide explains how to set up and use the service with your own design
 ### 2\) Create your own design
 
 * [Download](getting_started.md#download-project-config-from-service) the project config from the Livingdocs Service
-* Understand and modify the [project config](reference-docs/project_config/README.md)
+* Understand and modify the [project config](reference-docs/project-config/README.md)
   * modify the project settings
   * add contentTypes
   * define components and metadata in contentTypes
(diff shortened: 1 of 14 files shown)


gabrielhase:
The documentation is actually not missing, it's just that the service gives you much less possibilities, e.g. you will not have the possibility to ever interact with ES through the service, so documenting our ES setup would only be confusing for service consumers.

gabrielhase (approved):
👍

ajwild:
> The documentation is actually not missing, it's just that the service gives you much less possibilities, e.g. you will not have the possibility to ever interact with ES through the service, so documenting our ES setup would only be confusing for service consumers.

@gabrielhase does it mean that this section should be removed:
https://developers.livingdocs.io/reference-docs/project-config/content_types#public-api-config

gabrielhase:
> @gabrielhase does it mean that this section should be removed:
https://developers.livingdocs.io/reference-docs/project-config/content_types#public-api-config

I don't know, the link is a deadlink 😆 
If the publication index can be configured via the project-config it should be in the service docs. If it's only possible with code or instance config, then not.

PR #350, opened 2020-12-28
add video storage config

meinradandermatt:
# Motivation
Add documentation for video storage config

### DIFF
--- a/reference-docs/server-configuration/config.md
+++ b/reference-docs/server-configuration/config.md
@@ -355,6 +355,35 @@ processing: {
 }
 ```
 
+#### Videos
+Added in: `release-2021-02`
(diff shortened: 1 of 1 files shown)


reference-docs/server-configuration/config.md
```diff
@@ -355,6 +355,35 @@ processing: {
...
 ```
 
+#### Videos
+Added in: `release-2021-02`
+For videos it is necessary to have a specific configuration. The storage, publicUrl and the uploadRestriction must be set.
+
+```js
+  videos: {
+   // must be public accessible
+    publicUrl: 'http://livingdocs-images-dev.s3.amazonaws.com',
```
marcbachmann, line 365: Could you change that to `livingdocs-videos-development`. Just to communicate that a separate buckets should be used instead of the images bucket, which is usually private.

PR #352, opened 2021-01-13
fix(links): fix internal links by introducing reference section in SUMMARY to actually include all the linked files

benib:
## Motivation

GitBook requires all files that are linked to to be included in SUMMARY.md. Otherwise they are considered static files and are not actually interpreted as Pages so the link resolution somehow doesn't do it's job.

This resulted in many links pointing to the commit blob on Github instead of the docs.livingdocs.io URL.

This PR fixes that by:
- Introducing a `/reference` folder and section in the `SUMMARY.md` to hold files we don't want to link to in the navigation tree at another place
- change all the links to these files


reference/base_filter.md
```diff
@@ -7,9 +7,9 @@ Base filters are used in different places and have a common format to construct
 ## Filter Query Format
 
 There are different places, where one can define base filter:
-- `baseFilters` (used in [custom dashboards](../editor_settings.md))
-- `defaultQueries` (used in [default dashboards](../../editor-configuration/default_dashboard_filter.md#filter-config-properties-see-example))
-- `emptySearchQueries` (used in [default dashboards](../../editor-configuration/default_dashboard_filter.md#filter-config-properties-see-example))
+- `baseFilters` (used in [custom dashboards](/reference-docs/project-config/editor_settings.md))
```
DaRaFF, line 10: Does `reference-docs` exist or is this a fault and should be named `/reference` (there are other places too)

DaRaFF (approved):
Cool, nice that this will work now! (check my comment and fix if needed)

reference/base_filter.md
benib: it exists and is something else. `reference-docs` contains all the configuration stuff. `reference` is new and contains files we point to from various places but don't want to have in the main navigation. They are added at the very bottom of the `SUMMARY.md` because they have to in a kind of hidden subsection thingy.

PR #353, opened 2021-01-25
chore(docs): extend auth docs

mokimo:
add documentation for the new `openid-connect` auth flow


mokimo:
@marcbachmann is this fine?

PR #354, opened 2021-01-31
doc(embedded documents): add minimal embedded documents guide

benib:
## Motivation

With `release-2020-12` we shipped a first iteration of an `Editable Teasers` feature. There are some caveats that are mentioned in the docs. Nevertheless having a configuration example and a little introduction into what this feature means can be helpful to have in the docs.

This PR adds that minimal explanation and examples.

### DIFF
--- a/guides/embedded_documents.md
+++ b/guides/embedded_documents.md
@@ -1,0 +1,61 @@
+# Embedded Documents / Editable Teasers
+
+With `release-2020-12` a new possibility to resolve includes is available. This feature is still somewhat experimental and some usecases are not covered yet.
+
+You can return Livingdocs Document JSON from an [include resolve function](../evaluation-guide/intro.md).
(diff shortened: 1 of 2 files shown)


guides/embedded_documents.md
```diff
@@ -0,0 +1,61 @@
...
+      return renderTeaser({params, options, publicationApi, documentApi})
+    }
+  }
+}
+```
+
+## Caveats
+- `editableContent` has no effect when more than one component is returned.
+- The includes won't be resolved when you load the publication via Public API `api/v1/documents/:documentId/latestPublication`.
+- When requestion a web rendition of a publication, the components returned from the include resolver won't be part of rendered html for now.
```
DaRaFF, line 61: requesting

DaRaFF (approved):

PR #358, opened 2021-02-11
Small section on how to avoid cors

mokimo:
### Description
Adds a small section on how to avoid CORS

### DIFF
--- a/devops/self-hosting.md
+++ b/devops/self-hosting.md
@@ -47,15 +47,3 @@ Notable required configurations and our recommended best practices are described
 The preferred solution is to serve the livingdocs server instance on the same domain as the editor to prevent CORS requests and have a better security as we don't need to make the login cookies accessible on multiple domains. Please expose the livingdocs server instance on `/proxy/api` and then in the `livingdocs-editor` environment configs, configure the _host:_ `module.exports={api: {host: '/proxy/api'}}`
 
 Alternatively you can use _proxiedHost_ instead of _host_ to proxy to a dns name of an internal service that's not accessible from the internet. We'll automatically set up a http/websocket proxy on `/proxy/api`.
-
-
-**Request headers**
-
-A non-standard request header that is required is the **x-forwarded-prefix**. 
-
-As an example on how to configure that in a standard kuberentes ingress can be found here:
-https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#x-forwarded-prefix-header
-
-In this example the x-forwarded-prefix would look like this:
-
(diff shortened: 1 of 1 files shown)


mokimo:
@marcbachmann is this fine?

devops/self-hosting.md
```diff
@@ -37,3 +37,25 @@ Both the containers for server and editor are stateless. The elasticsearch and p
...
+**Avoiding CORS**
+
+The preferred solution is to serve the livingdocs server instance on the same domain as the editor to prevent CORS requests and have a better security as we don't need to make the login cookies accessible on multiple domains. Please expose the livingdocs server instance on `/proxy/api` and then in the `livingdocs-editor` environment configs, configure the _host:_ `module.exports={api: {host: '/proxy/api'}}`
+
+Alternatively you can use _proxiedHost_ instead of _host_ to proxy to a dns name of an internal service that's not accessible from the internet. We'll automatically set up a http/websocket proxy on `/proxy/api`.
+
+
+**Request headers**
+
+A non-standard request header that is required is the **x-forwarded-prefix**.
```
marcbachmann, line 54: We should change the wording here. It is not required as long as we use `/proxy/api`.
We should also add a note about rewriting the path from `/proxy/api` to `/`. Maybe we could also mention envoy?

But in the default setup everything will work now if a customer uses `/proxy`.

devops/self-hosting.md
```diff
@@ -37,3 +37,25 @@ Both the containers for server and editor are stateless. The elasticsearch and p
...
+The preferred solution is to serve the livingdocs server instance on the same domain as the editor to prevent CORS requests and have a better security as we don't need to make the login cookies accessible on multiple domains. Please expose the livingdocs server instance on `/proxy/api` and then in the `livingdocs-editor` environment configs, configure the _host:_ `module.exports={api: {host: '/proxy/api'}}`
+
+Alternatively you can use _proxiedHost_ instead of _host_ to proxy to a dns name of an internal service that's not accessible from the internet. We'll automatically set up a http/websocket proxy on `/proxy/api`.
+
+
+**Request headers**
+
+A non-standard request header that is required is the **x-forwarded-prefix**. 
+
+As an example on how to configure that in a standard kuberentes ingress can be found here:
```
marcbachmann, line 56: That annotation mentioned is not a standard kubernetes ingress (It's for nginx). The standard declaration doesn't support customizations and because of that most integrations use a hack and use annotations as configs.

devops/self-hosting.md
mokimo: I think we could skip on the request headers and consult customers if they run into issues. They shouldn't if they use the standard setup we advise in the `avoiding cors` section

PR #361, opened 2021-02-23
Print documentation

gabrielhase:


gabrielhase:
@nephridium This might be interesting for you as well (available with the next release). If you have a look at it, could you also quickly look over the "Implementation Details" section? That section is very old and I don't know if the API has changed in the meantime.

nephridium:
@gabrielhase LGTM, can't see any glaring changes we did to the API for the NZZ use case, I think it should generally work with configs like these. (We have some custom copy behavior here and there, but such things are very much up to the customer.)

PR #364, opened 2021-02-24
doc(media library): add doc for computed media library dashboards

benib:
Related PR: https://github.com/livingdocsIO/livingdocs-editor/pull/4180

### DIFF
--- a/reference-docs/project-config/editor_settings.md
+++ b/reference-docs/project-config/editor_settings.md
@@ -99,9 +99,11 @@ An example:
       {metadataPropertyName: 'source', directiveName: 'source'},
       {metadataPropertyName: 'caption', directiveName: 'caption'}
     ],
+    // deprected with release-2021-03, move filters to a mediaType
     dashboard: {
       displayFilters: ['timeRange']
     },
+    // deprected with release-2021-03, move filters to a mediaType
     editorSelection: {
       displayFilters: ['timeRange']
     },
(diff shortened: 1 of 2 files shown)


reference-docs/project-config/editor_settings.md
```diff
@@ -99,9 +99,11 @@ An example:
       {metadataPropertyName: 'source', directiveName: 'source'},
       {metadataPropertyName: 'caption', directiveName: 'caption'}
     ],
+    // deprected with release-2021-03, move filters to a mediaType
```
DaRaFF, line 102: I'm not sure if it is a good idea to deprecate such configs. In my opinion it's easier to force the customers to move the configs immediately and make the config invalid if they use the old configs. Moving a config is not that hard and if they do it now, it's better for us, because we can already remove backward compatible checks and make 2 steps. For more complicated stuff I agree it makes sense to add a deprecation.

reference-docs/project-config/editor_settings.md
```diff
@@ -161,11 +163,15 @@ Type of the dashboard, one of these: `dashboard`, `kanbanBoard`, `taskBoard`
 
 #### displayFilters
 
-[Display Filters](/reference/display_filter.md) are filters that the user can set in the UI (below the search input)
+[Display Filters](/reference/display_filter.md) are filters that the user can set in the UI (below the search input).
+
+With `release-2021-03` the `displayFilters` for the Media Library Dashboards are defined on the [Media Type](/reference-docs/project-config/media_types.md).
```
DaRaFF, line 168: Maybe it's a good idea to describe the default mediaType config in `/reference-docs/project-config/media_types.md`, so that the customers have an easier life to migrate.

DaRaFF (approved):

reference-docs/project-config/editor_settings.md
benib: that is also done over there.

reference-docs/project-config/editor_settings.md
benib: I went with deprecation. Was easy to support. We can kill at the next iteration.

PR #372, opened 2021-03-11
doc(references): Add references content model

ajwild:
I'm not sure if this should be listed as a content model, or have its own section directly under "Reference Docs".

Examples taken from:
- https://github.com/livingdocsIO/livingdocs-server/pull/3341
- https://github.com/livingdocsIO/livingdocs-server/pull/3383

### DIFF
--- a/reference-docs/content-model/references.md
+++ b/reference-docs/content-model/references.md
@@ -1,0 +1,127 @@
+# References
+
+References are stored for documents and media library entries. They indicate links from the document or entry to other content.
+
+## Examples
+
+### Within document content
+
+References extracted from the content of a document provide details of the item they are referencing (`id` and `type`) along with contextual information to help locate the reference within the document.
+
+#### Include directive params:
+
+```
+{
+  type: 'document',
+  id: '466',
+  location: 'include-directive',
+  componentId: 'doc-478e2',
+  componentName: 'social-media-embed',
+  directiveName: 'embed',
+  serviceName: 'twitter-include',
+  propertyName: 'document'
+}
+```
+
+#### Include directive override:
+
+```
+{
+  type: 'document',
+  id: '223',
+  location: 'include-directive-override',
+  componentId: 'doc-472ka3',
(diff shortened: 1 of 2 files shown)


DaRaFF (approved):
I like how it is currently

reference-docs/content-model/references.md
```diff
@@ -0,0 +1,127 @@
+# References
+
+References are stored for documents and media library entries. They indicate links from the document or entry to other content.
```
benib, line 3: I'd love to see pointers to the Public API explaining how I can get to see these references.

benib (changes requested):

reference-docs/content-model/references.md
ajwild: I didn't want to add too much info about the Public API into this content-model section, but a pointer is a good idea. Should I add a text description of how to find the public API docs and the section the references endpoints are in, or include a link to localhost?

peyerluk: We just link to edit.livingdocs.io when we refer to the API documentation.

ajwild: Done. Now that I think about it, I find it a bit strange that I've exposed the `references` property on 4 endpoints outside the Public API beta paths. I've deliberately not mentioned them in the docs here. We'll have to have a think about which endpoints should include references when we move from beta to v1, and either rethink how we select fields, or remove the references once they've been loaded from the database (not ideal).

For info, it's the copy, transform, and republish endpoints for documents, as they return the document without any transforms, and the more commonly used `/media-library/:mediaLibraryEntryId`. There's also the `/media-library?reference=image:FxZE7difmfvW` query which I should probably not have done, as it doesn't return any references :facepalm:

benib (approved):

reference-docs/content-model/references.md
benib: I don't fully understand this yet. But I guess you will act on it to come to a conclusion?

PR #386, opened 2021-04-09
Add doc for storage prefix

DaRaFF:


marcbachmann (approved):
Do we backport that into `release-2021-03`?
Or should we already create a `release-2021-05` branch in here and not merge to master?

DaRaFF:
> Do we backport that into release-2021-03?
> Or should we already create a release-2021-05 branch in here and not merge to master?

In my view, we should continue with the strategy to just write when something was introduced with `introduced in release-2021-06`. When we have bigger changes in structure, we have to think about how we can solve that in the future.

content/enterprise/reference-docs/server-configuration/_index.md
```diff
@@ -295,6 +295,7 @@ designs: {
     publicUrl: 'http://livingdocs-designs-dev.s3.amazonaws.com',
     storage: {
       strategy: 's3',
+      prefix: 'images/' // optional, the storage key will be prefixed (added in release-2021-06)
```
marcbachmann, line 298:
```suggestion
      prefix: 'images/' // optional, the storage key will be prefixed ({{< added-in release-2021-06 >}})
```

marcbachmann (approved):

PR #390, opened 2021-04-13
fix(import-api): update import docs

mokimo:
Server PR: https://github.com/livingdocsIO/livingdocs-server/pull/3569

### Changelog
Add `publicationDate` docs and some more numbers on estimations of how fast imports currently are.

### DIFF
--- a/content/enterprise/guides/integrations/import-legacy-system-documents.md
+++ b/content/enterprise/guides/integrations/import-legacy-system-documents.md
@@ -8,14 +8,13 @@ menus:
 
 
 ## Foreword and import statistics
-As imports from old systems can contain millions of documents, it's important to correctly estimate the time needed to complete an import.
+Importing millions of documents from legacy systems into Livingdocs takes time. We observed these numbers:
+- 50k articles per hour
+- 100k - 300k images per hour
 
-As a rough number, you can assume to import ~ 50.000 articles and ~ 100.000 - 300.000 images per hour. Without the necessity to import images, you could import a lot more articles per hour.
+During these observations, memory usage was around ~4GB or RAM und 25Mbps of inbout and outbound bandwidth was used.
 
-This scenario assumes ~4 GB of RAM usage and a bandwidth of 25Mpbs inbound and outbound. 
-
-Of course, the Infrastructure can be scaled to support faster imports if that's necessary.
-
+If no images are imported, a lot more documents could be imported. 
 
 ## Custom document IDs
 
--- a/content/enterprise/guides/integrations/import-legacy-system-documents.md
+++ b/content/enterprise/guides/integrations/import-legacy-system-documents.md
@@ -6,9 +6,20 @@ menus:
     parent: Integrations
 ---
 
-{{< added-in release-2020-10 >}}
 
-## Motivation
+## Foreword and import statistics
+As imports from old systems can contain millions of documents, it's important to correctly estimate the time needed to complete an import.
+
+As a rough number, you can assume to import ~ 50.000 articles and ~ 100.000 - 300.000 images per hour. Without the necessity to import images, you could import a lot more articles per hour.
+
+This scenario assumes ~4 GB of RAM usage and a bandwidth of 25Mpbs inbound and outbound. 
+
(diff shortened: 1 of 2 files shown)


content/enterprise/guides/integrations/import-legacy-system-documents.md
```diff
@@ -6,9 +6,20 @@ menus:
...
 
-## Motivation
+## Foreword and import statistics
+As imports from old systems can contain millions of documents, it's important to correctly estimate the time needed to complete an import.
+
+As a rough number, you can assume to import ~ 50.000 articles and ~ 100.000 - 300.000 images per hour. Without the necessity to import images, you could import a lot more articles per hour.
+
+This scenario assumes ~4 GB of RAM usage and a bandwidth of 25Mpbs inbound and outbound. 
+
+Of course, the Infrastructure can be scaled to support faster imports if that's necessary.
```
benib, line 17: This is not super clear to me. It gives me some rough numbers and talks about memory and bandwidth usage when these numbers where observed. That's good 👍 
Then it tells me the infrastructure can be scaled to support faster imports. To be complete, this should also tell me what the bottleneck is that would need to be scaled.

Just a suggestion:

```
Importing millions of documents from legacy systems into Livingdocs takes time. It is mostly (?Database IO?) bound. With a single database node setup (improve wording, link to docs) we observed these numbers:
- 50k articles per hour
- 100k - 300k images per hour

During these observations, memory usage was around ~4GB or RAM und 25Mbps of inbout and outbound bandwidth was used.

If no images are imported, a lot more documents could be imported. 

If you want to scale the infrastructure to handle more imports/time, you [explain what first to do]
```

content/enterprise/guides/integrations/import-legacy-system-documents.md
```diff
@@ -65,3 +76,18 @@ curl -k -X POST "https://edit.livingdocs.io/proxy/api/v1/import/documents" \
 }
 EOF
 ```
+
+## Custom publication dates
+
+{{< added-in release-2021-03 >}}
+
+When importing articles from legacy systems, you should be setting the `publicationDate`. The `publicationDate` can be found in the [public api documentation](https://edit.livingdocs.io/public-api) or [import api reference documentation]({{< ref "../../reference-docs/server/import-api.md" >}}).
```
benib, line 84: I'd add a shortened example of where/how I set the `publicationDate` during import and in what format I need to provide it.

content/enterprise/reference-docs/server-api/import_api.md
```diff
@@ -26,6 +28,7 @@ const importLog = await import({importJob, rawDocument, shouldCreateNew, updateC
 * `channelid`: A `number` containing the Id of the Channel you want the Document to import to.
 * `contentType`: A `string`, the desired `contentType` of your Document.
 * `title`: A `string`, the title of your Document.
+* `publicationDate` A `string` containing an ISO date of when an article was published (`autoPublish` flag needs to be set as well)
```
benib, line 31: I assume this is `ISO 8601`?

benib (approved):

content/enterprise/guides/integrations/import-legacy-system-documents.md
mokimo: Not 100% sure either what the bottleneck is, or what would need to be scaled - we could remove that paragraph instead of making guesses and just mention the import numbers. 

I removed the unclear parts for now

content/enterprise/guides/integrations/import-legacy-system-documents.md
mokimo: There's one in the public-api documentation at https://0.0.0.0:9000/public-api and I'll backport it so we can see it on the march release on https://edit.livingdocs.io/public-api too

That documentation here rather focuses on concepts, so I'd keep that example in the public-api docs

content/enterprise/reference-docs/server-api/import_api.md
mokimo: Yes, updated that with a note

PR #394, opened 2021-04-30
elasticIndex docs and more

marcbachmann:
### Changelog
- 🎁 Add redis configuration documentation
- 🎁 Add elasticIndex.concurrency, batchSize and maxCpu documentation
- 🎁  We now list the whole configuration in one block and then reference to the details using links
image


content/enterprise/operations/external-services.md
```diff
@@ -1,52 +1,68 @@
 ---
-title: External Services
+title: Software Requirements and Dependencies
```
DaRaFF, line 2: @marcbachmann is this site available in the docs? I can't find it.

marcbachmann: not in the menu at the moment. ee should put it into the operations section

PR #396, opened 2021-05-04
Add guide for image storage + delivery proxy

DaRaFF:


content/enterprise/guides/integrations/image-upload-delivery-service.md
```diff
@@ -0,0 +1,131 @@
...
+    url: 'http://localhost:4444'
+  }
+},
+
+documents: {
+  // info: when imageProxy: enabled, then you must define
+  //       selectedImageService: 'liImageProxy'
+  selectedImageService: 'liImageProxy',
+  imageServices: {
+    liImageProxy: {
```
marcbachmann, line 44: Please just directly use imgix and serve images from the custom http server.
That configuration is way too complicated and confusing.

content/enterprise/guides/integrations/image-upload-delivery-service.md
```diff
@@ -0,0 +1,131 @@
+---
+title: Image Storage & Delivery Proxy
+tags: [guides, integrations]
+menus:
+  guides:
+    parent: Integrations
+---
+
+## Integration Guide for your own Image Storage and Image Delivery with an Imgix API
```
marcbachmann, line 9: Please use shorter titles if possible.
`own Image Storage` is misleading as that would also be supported using https://min.io and https://imgproxy.net

-> Integrate a proprietary HTTP-based Image Storage

content/enterprise/guides/integrations/image-upload-delivery-service.md
```diff
@@ -0,0 +1,131 @@
...
+
+Most customers just use [Image Services]({{< ref "/enterprise/guides/image-services#overview" >}}), because they usually support everything you need. If you want to implement your own service (described in this guide), please first get in touch with us to discuss the advantages/disadvatages.
+
+In the following examples you will see the most simple possible service and configuration to support your own upload, storage and delivery.
+
+In a first step you need to configure your own service in the server-config of Livingdocs. Most of the values in the example below can't be changed, because otherwise the setup does not work anymore.
+These values are changeable:
```
peyerluk, line 16: I think we should try to be able to explain the values in the config further below in the config instead of describing up here which values can be changed. Currently it makes it sound a bit more complicated than it is.

PR #398, opened 2021-05-11
Update openid-connect integration examples

mokimo:
Add an openid-connect example for google

### DIFF
--- a/content/enterprise/reference-docs/server-configuration/single_sign-on.md
+++ b/content/enterprise/reference-docs/server-configuration/single_sign-on.md
@@ -28,15 +28,40 @@ In the following example we set three different providers: Github, Google and Fa
         config: {
           clientId: '',
           clientSecret: '',
+          // jwtContentFromSSOProvider is a openid-connect standard claim, for available properties see:
+          // https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
           extractGroupNames: async (jwtContentFromSSOProvider, groupsOfProject) => {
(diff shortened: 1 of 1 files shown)


content/enterprise/reference-docs/server-configuration/single_sign-on.md
```diff
@@ -28,17 +28,42 @@ In the following example we set three different providers: Github, Google and Fa
...
           clientSecret: '',
+          // jwtContentFromSSOProvider is a openid-connect standard claim, for available properties see:
+          // https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
           extractGroupNames: async (jwtContentFromSSOProvider, groupsOfProject) => {
             return ['owners']
           },
-          // defaultProjectId: 1,
-          defaultProjectHandle: 'enterprise-web-print'
+          // User will be created and logged into this project (enterprise use-case)
+          defaultProjectHandle: 'enterprise-web-print' // alternative 'defaultProjectId: 1'
```
DaRaFF, line 37: wow, is there any use case where `defaultProjectId` make sense?

DaRaFF (approved):

content/enterprise/reference-docs/server-configuration/single_sign-on.md
mokimo: We're inconsistent about the use of ID/Handle and pretty much always support both, so supporting both here is the most consistent solution.

PR #404, opened 2021-05-27
doc(import): Add copy documents between projects page

ajwild:
I'm not sure how this should be named:
- Copy documents between projects
- Cross-project document import
- Import document and media


peyerluk:
I like 'Copy documents between projects' the best.

content/enterprise/guides/copy-documents/copy-documents-between-projects.md
```diff
@@ -0,0 +1,217 @@
...
+---
+## Overview
+
+This feature allows you to drag-and-drop a document from one project to another. It is limited to projects on the same server and in the same environment. It is best suited to copying documents between projects with the same design.
+
+## Configuration
+
+### Server config
+
+A new property, `customerId`, is required in the server config to help identify the source and destination servers. The value is a string which will typically be the same across all of your environments.
```
peyerluk, line 14: The `customerId` should also be added to the reference documentation.

content/enterprise/guides/copy-documents/copy-documents-between-projects.md
```diff
@@ -0,0 +1,217 @@
...
+
+```js
+{
+  ...serverConfig,
+  customerId: 'livingdocs'
```
peyerluk, line 19: We should take the customers perspective and use a different example than 'livingdocs'. Something like 'daily-planet' or 'acme-corp'...

content/enterprise/guides/copy-documents/copy-documents-between-projects.md
peyerluk: Also we could mention that currently we only support copying documents with the same 'customerId' and 'environment' which effectively means the same server.

peyerluk, line 14: > The value is a string which will typically be the same across all of your environments.

We could extend this with: 'If you use multiple Livingdocs servers for different projects you should give them different `customerIds`'

content/enterprise/guides/copy-documents/copy-documents-between-projects.md
```diff
@@ -0,0 +1,217 @@
...
+}
+```
+
+### Project config
+
+The project config requires two new properties, one which allows documents to be exported to specific projects, and another which configures how to import documents from other projects.
```
peyerluk, line 25: The new properties should also be added to the reference docs.

content/enterprise/guides/copy-documents/copy-documents-between-projects.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Feature API
+
+You may want to write some custom functionality or a script which copies documents between projects. For this reason we expose the `importDocumentAndMedia()` function on the `li-import` feature API.
+
+#### Request
+
+When calling the `importDocumentAndMedia()` function you need to provide the (destination) `projectId` and (destination) `userId`, along with information about the source document and server. The source `documentId` and `projectId` are mandatory, while the `channelId` is optional. The source server data, `customerId` and `environment`, should be taken from the server config. If you have already loaded the `documentVersion` from the source project you can pass it through as `sourceDocumentVersion` which saves on a database query. There is also a `verifyDocumentCreatePermissions` property, which allows you to pass a function to verify that the user has permission to create the document. This hook receives the new `documentVersion` after it has been created and transformed. The hook must throw and error to prevent the document from being commited to the database.
```
peyerluk, line 92: maybe: ...provide the (destination) `projectId` and a `userId` of the user who will become the owner of the copied documents.

DaRaFF (approved):
n1 👍

peyerluk (approved):

PR #405, opened 2021-05-27
chore(npm): Make start script cross-platform

ajwild:
This ends up adding quite a few packages. Should I look for another way?

The old command would open the browser before the server started, requiring a page refresh. Now it waits for the port to respond before opening the browser.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "livingdocs",
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
@@ -14,6 +15,9 @@
         "glob": "^7.1.6",
         "htmlparser2": "^6.0.0",
         "lunr": "^2.3.9"
+      },
+      "devDependencies": {
+        "opener": "^1.5.2"
       }
     },
     "node_modules/balanced-match": {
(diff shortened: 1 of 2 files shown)


marcbachmann:
> The old command would open the browser before the server started, requiring a page refresh.

Chrome does this automatically for me 😁 

Should be fine adding those dependencies.
But the wait-for-localhost-cli dependency tree is huge. This bash script would do the same, but won't have windows compatiblity:

```
wait-for-service localhost 1313 && open-cli http://localhost:1313
```

content of `wait-for-service`
```bash
#!/bin/sh
set -e
TIMEOUT=$((TIMEOUT))
if [ "$TIMEOUT" == "" ] || [ "$TIMEOUT" == "0" ]; then TIMEOUT=300; fi

function after () {
  local REMAINING=$1
  echo "after $((TIMEOUT - REMAINING))s"
}

function wait_for_service () {
  local HOST="$1"
  local PORT="$2"
  local REMAINING=$TIMEOUT
  >&2 printf "wait-for: Waiting for TCP connection to $HOST:$PORT..."
  until nc -zw 1 $HOST $PORT 2>/dev/null
  do
    if [ "$REMAINING" -lt 1 ]; then
      >&2 printf "TIMEOUT $(after $REMAINING)\n"
      >&2 echo "wait-for-services: Connection to $HOST:$PORT failed"
      exit 1
    fi
    sleep 1
    >&2 printf .
    REMAINING=$((REMAINING - 1))
  done

  >&2 printf "OK $(after $REMAINING)\n"
}

wait_for_service $1 $2
```

marcbachmann:
https://github.com/Chris927/wait-for-host wouldn't have any dependencies.

ajwild:
We could use:
```
  "devDependencies": {
    "npm-run-all": "^4.1.5",
    "opener": "^1.5.2",
    "wait-on": "^5.3.0"
  },
  "scripts": {
    "build": "./bin/build.sh",
    "open": "run-s open:wait open:browser",
    "open:wait": "wait-on http://localhost:1313",
    "open:browser": "opener http://localhost:1313",
    "server": "hugo server",
    "start": "run-p server open"
  },
```
but it still makes package-lock.json ~1910 lines, up from ~360 lines.

This latest update only adds ~20 lines, but I'll need to refresh Firefox :sob:

Why do we need `exec`?

marcbachmann:
`exec` is not really needed. It's just there to replace the shell process of npm with the process passed to exec.

PR #412, opened 2021-07-04
feat: Add revoke media page and events

ajwild:
Relations:
  - Issue: https://3.basecamp.com/4910529/buckets/19620057/documents/3884808654
  - Related PR's:
    - https://github.com/livingdocsIO/livingdocs-server/pull/3809
    - https://github.com/livingdocsIO/livingdocs-editor/pull/4516

Remaining questions:
1. Should we make a media library section in the guides, split the existing content and add this as another page there?
2. Alternatively, could this be a workflow where we add suggestions about reviewing references and replacing the image?
3. Should I add some more examples? If so, what?

### DIFF
--- a/content/enterprise/guides/revoke-media.md
+++ b/content/enterprise/guides/revoke-media.md
@@ -1,0 +1,47 @@
+---
+title: Revoke Media
+description: Handle the revoke event to complete the removal of media
+menu: guides
+tags: [guides]
+---
+
+The Livingdocs server handles the basic functionality of revoking media - removing the asset(s) from storage, removing the media library entry from the dashboard, and updating the media library entry state.
+It is likely that each customer will have their own requirements for additional steps required to complete the revoke process based on their server configuration and editorial processes.
+These steps may include:
+- clearing caches
+- regenerating pages
+- replacing the media with a placeholder
+- sending an email notification
+
+To be as flexible as possible we have provided a server event and a webhook, both named `mediaLibraryEntry.revoke`, which the downstream server can subscribe to.
+
+## Example
+
+### imgix cache purge
+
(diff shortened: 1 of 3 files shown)


content/enterprise/guides/revoke-media.md
```diff
@@ -0,0 +1,42 @@
...
+To be as flexible as possible we have provided a server event and a webhook, both named `mediaLibraryEntry.revoke`, which the downstream server can subscribe to.
+
+## Example
+
+### imgix cache purge
+
+As is common for many content delivery networks (CDN), imgix will cache your source image for up to one year on its own servers.
+After the Livingdocs server deletes the asset from storage it could still take days, weeks or even months before the image is removed from the CDN.
+In order to avoid this delay it is important to instruct imgix to clear it's cache for the asset and serve the latest version, or in this case a missing/broken image.
+Below is an example of simple purge request sent to imgix using the `axios` library and an API key for authorization:
```
DaRaFF, line 25:
```suggestion
Below is an example of a simple purge request sent to Imgix using the `axios` library and an API key for authorization:
```

DaRaFF (approved):

content/enterprise/guides/revoke-media.md
```diff
@@ -0,0 +1,42 @@
...
+
+```js
+liServer.events.subscribe('mediaLibraryEntry.revoke', async (event, {mediaLibraryEntry}) => {
```
marcbachmann, line 28: Don't use async with node eventemitters if you don't handle all the errors

content/enterprise/guides/revoke-media.md
marcbachmann: imgix lower case should be good as imgix write it like themselves and we apply the same style in our whole documentation.

PR #420, opened 2021-09-07
chore: More reasonable sample value for maxArea

romankaravia:
Setting a non-null `maxArea` leads to behavior that is probably not desired in most cases in combination with "free" cropping (without fixed aspect ratio), so I suggest it to `null` in the code sample.

Case in point: T-Online were confused by this behavior 👇

https://user-images.githubusercontent.com/47303530/132364105-d2370395-73ed-4ec5-a67f-dcc9b102d409.mov


benib (approved):
I believe you could even remove it completely from the docs.

PR #445, opened 2021-12-20
Instant Publish Process

DaRaFF:


DaRaFF:
@marcbachmann I already merge it, because I need the doc for customers and also want to make new PRs based on this one. You're feedback is still highly welcome.

PR #448, opened 2021-12-23
Handsontable Include and iframe modal

JSchenk8:
- Here are a draft of the docs for Handsontable includes. 
- Because you need to use modules to run a handsontable instance, I have left it up to the user to decide how to render a handsontable, but example file set ups are free to see on the handsontable demo, linked in the docs.


meinradandermatt (changes requested):
I think you should add more introduction that the devs know what different things they have to do. Specially to run handsontable as a seperate application. And at the handsontable itself it should be also described what steps are needed.

content/guides/integrations/handsontable/index.md
```diff
@@ -0,0 +1,333 @@
...
+        return `<th :key=${index}>${item}</th>`
+      }
+    })
+    return `<tr :key=${outerIndex}>${valueArrayHTML.join('')}</tr>`
+  }).join('')
+}
+```
+## Sending and receiving data with Livingdocs
+
+The function for the iframe modal to send messages is already written in the Editor, it can be found in the includes index file.
```
meinradandermatt, line 102: I would just say here that the editor is waiting on a config request and is sending then the config.

meinradandermatt (changes requested):

meinradandermatt (approved):

meinradandermatt (approved):

PR #456, opened 2022-01-11
fix: Update date of next release

romankaravia:


DaRaFF:
This is not necessary, I did that already in my open branch with all release updates. But thanks for watching :-)

PR #457, opened 2022-01-16
feat: Move incoming references public api endpoints out of beta

ajwild:
Relations:
  - Related PR's: https://github.com/livingdocsIO/livingdocs-server/pull/4171

# Motivation

Moving the four incoming references endpoints out of beta for release-2022-03.


ajwild:
@DaRaFF when would this be merged? When `release-2022-03` is officially released? Or do we need to introduce an "added-in" label to the Public API template?

DaRaFF (approved):
I would merge it now but put an `added-in` label.

svenlivingdocs (approved):
Looks good, thx!
Will style the release line later down the road.
Cheers!

PR #459, opened 2022-01-21
Vue metadata plugins

benib:
Docs for: https://github.com/livingdocsIO/livingdocs-editor/pull/4963

Merge for March Release.


DaRaFF (approved):
I fixed the format in the plugin list. It was formatted wrong (stuff on the same line, no js formatting, and so on). I think we can merge it like that as a first step, but we have to continue to work on. We need more screenhots, config examples and description of the purpose of the plugin

PR #460, opened 2022-01-21
fix: Add 4 missing callback support removals in 2021-09 release notes

romankaravia:
The list of breaking changes is now the same as the list of referenced PRs further down.

(discovered while migrating heute.at to 2021-09)


DaRaFF (approved):
thx!

PR #461, opened 2022-01-21
WIP: feat: Add video includes guide

ajwild:
I'll complete this once https://github.com/livingdocsIO/livingdocs-server/pull/4197 is approved and merged.


content/guides/media-library/video-includes/index.md
```diff
@@ -0,0 +1,276 @@
...
+  const documentsConfig = liServer.features.config('li-documents')
+  const imageService = framework.imageServices.get(documentsConfig.selectedImageService)
+
+  return {
+    name: 'video-service',
+    paramsSchema: [
+      {
+        handle: 'video-param',
+        type: 'li-video-reference',
+        // Preloading is only neccessary if you need the asset or some metedata,
```
DaRaFF, line 69: metadata

content/guides/media-library/video-includes/index.md
```diff
@@ -0,0 +1,276 @@
...
+                  const errorEl = document.querySelector('#${randomId} .error')
+                  errorEl.style.display = 'flex'
+                }, 10000)
+              })()
+            </script>
+          `,
+          dependencies: {
+            js: [
+              {
+                src: 'https://unpkg.com/video.js@7.17.0/dist/video.js',
```
DaRaFF, line 196: Is it a good idea to refer to an open-source project with no service guarantee? //cc @marcbachmann

content/guides/media-library/video-includes/index.md
```diff
@@ -0,0 +1,276 @@
...
+// components/video.js
+
+module.exports = {
+  name: 'video-include',
+  label: 'Video',
+  iconUrl: `https://livingdocsio.github.io/livingdocs-design-assets/docs/icons/component-icons/icon_component_video.svg`,
+  directives: [{
+    name: 'video-directive',
+    type: 'include',
+    service: 'video-service'
```
DaRaFF, line 246: Should we define core includes like metadata plugins in the future? Otherwise, customers have to copy/paste code for default behavior. We did it for metadata plugins, why not do it for includes? //cc @benib @peyerluk

DaRaFF (approved):
We should wait for feedback from @marcbachmann regarding unpkg, but otherwise, it looks good!

content/guides/media-library/video-includes/index.md
ajwild: The video-service is not a core service. Only the `paramsSchema` form extension to add support for videos and poster images was added to the core (`type: 'li-video-reference'` within `paramsSchema`). However, your comment is relevant to the recently merged oembed service which is our first core include service. At the moment it's registered with `service: 'li-oembed-service'`.

content/guides/media-library/video-includes/index.md
ajwild: Video.js is available through [cdnjs](https://cdnjs.com/libraries/video.js), which covers two of the five dependencies, but the theme and the plugin aren't hosted on a (public) CDN. For a code example I don't think we should host this ourselves, and the demo won't work without the plugin, so I just picked the same CDN for all dependencies for consistency. Maybe I can add a comment somewhere...

content/operations/releases/master/master.md
```diff
@@ -104,10 +104,10 @@ These are the release notes of the upcoming release (pull requests merged to mas
 
 ## Video Includes
 
-TODO: add a description + documentation
+Video includes provide a more flexible approach to embedding videos within a document than using the standard media library entry directly. The media library approach does not have an integrated solution for video transcoding, so it would only be possible to render the original video file to HTML. On the other hand, using an include service allows you to hook into an external service or data store which could provide the transcoded assets you require.
```
marcbachmann, line 107: Unrelated to this, but might be useful later: imgix announced video transcoding in december: https://blog.imgix.com/2021/12/13/streaming-video-part-1-businses-value

At the moment it's only available for enterprise customers, unless we contact them.

content/guides/media-library/video-includes/index.md
```diff
@@ -0,0 +1,276 @@
...
+                  label="${asset.label}"
+                  ${asset.selected ? 'selected="true"' : ''}
+                >
+              `).join('')}
+              <div class="status-container">
+                <div class="status-content loading">Loading...</div>
+                <div class="status-content error">Unable to load video player</div>
+              </div>
+            </video-js>
+            <script>
```
marcbachmann, line 156: Can we move that whole logic into the component or a js file of the design itself?
Also the assets

ajwild: It's easier this way when I want to make sure things are scoped by `randomId`, and to set at the correct aspect ratio. I guess I could use a data attribute for the aspect ratio, and maybe do some sort of `closest()` call to find the relevant video to load.

I don't think we have a JS file in the design, and I don't know where we'd host it and how we'd maintain it. However, I could add the assets to the design.

content/operations/releases/master/master.md
peyerluk: > imgix announced video transcoding

That sounds useful! We should try this out at some point.

content/guides/media-library/video-includes/index.md
marcbachmann: At the moment we can keep it as is, but this would be simple:

```js
<script>someDesignScripts.initializeVideo(${JSON.stringify(randomId)})</script>
```

marcbachmann: or we finally integrate some design hooks which could be similar.
directly in the design
```
if (typeof window.liEditor) {
  liEditor.onComponentRender((c) => {
    if (c.component !== 'video-include') return
    someDesignScripts.initializeVideoComponent(c)
  })
} else {
  for (const video of document.querySelectorAll('video-js')) {
    someDesignScripts.initializeVideoComponent(component)
  }
}
```

@peyerluk

marcbachmann: But I think you could move the whole script into the dependencies. That way they can get deduplicated.

That could be cleaner
```
dependencies: [{
  src: `
    window.renderVideoIncludeComponent = function (id) {
  
    }
  `
  namespace: 'includes.video'
},
// And one entry without namespace, to prevent deduplication?
{
  src: `renderVideoincludeComponent(${JSON.stringify(randomId)})`
}]

```

content/guides/media-library/video-includes/index.md
marcbachmann: I'm fine with using unpkg

marcbachmann (approved):

content/guides/media-library/video-includes/index.md
ajwild: @marcbachmann I've gone with the `dependencies` change as you mentioned in the comment above. FYI it seems as though `inline: true` is not necessary as this is computed based on whether `src` or `code` has been defined.

PR #472, opened 2022-03-02
feat: Add Retresco setup guide

ajwild:
Relations:
  - Related PR's:
    - https://github.com/livingdocsIO/livingdocs-server/pull/4215
    - https://github.com/livingdocsIO/livingdocs-editor/pull/5013

I think it's mostly the text that needs checking - especially the first paragraph. I've run through the guide and setup Retresco for BLZ as a test and the configs worked fine with the latest release-2022-03.


content/guides/integrations/retresco.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+### Project Secret
+
+Finally, a secret must be added for the relevant `<project-handle>`. The `<secret-name>` should match the value of the `password.$secretRef.name` property in the Retresco integration's project config object. The `<password>` should be the password provided by Retresco used to access the API. Within a development or test environment it is also possible to seed this value.
+
+```bash
+npx livingdocs-server secret-add --project="<project-handle>" --name="<secret-name>" --value="<password>"
+```
```
benib, line 86: If we already have some kind of general secret management guide, this should be referenced here. If we don't, I believe it would be a good addition to our documentation. Then we don't have to repeat too much of this with every feature using the secrets. I'd like to leave this to @livingdocsIO/team-developer-experience to lead the effort of documenting this.

benib (approved):
It would be nice to see a link from the metadata plugins list entry for `li-retresco` to this guide. But this one here should be merged and another follow up could extend this.

content/guides/integrations/retresco.md
DaRaFF: Yes true, I added a task for that (but this PR can already be merged).

ajwild: I agree, it would be nice to have a guide about secrets and secret seeding, then all references to it can be standardised with a short sentence and a link.

I've added two links from other parts of the docs to this guide.

PR #477, opened 2022-03-07
Add 2022-03 release notes for table dashboard

romankaravia:


content/operations/releases/release-2022-03/release-2022-03.md
```diff
@@ -102,10 +102,23 @@ References:
 
 ## Table Dashboards
 
-TODO@roman: add description + documentation
+Table dashboards are a new flexible type of dashboards where individual table cells can be configured.
```
benib, line 105: I wonder if this should be `...individual column` can be configured.`. But cell is also fine.

benib (approved):

PR #478, opened 2022-03-15
Metadata plugins

benib:
* li-document-reference
* li-video-reference


DaRaFF (changes requested):
Please also add a screenshot and a config example.

DaRaFF (approved):

PR #483, opened 2022-03-20
replace sso guide

meinradandermatt:
# Motivation
Our SSO guide was had wrong links and was out of date. So I added a new example for AzureAD which is the most common use-case for our customers.


content/guides/authentication/azuread-login.md
```diff
@@ -0,0 +1,112 @@
+---
```
DaRaFF, line 1: please rename that file to `azure-ad-login`... I was still a bit tired and read "azu..... read"... hä what "read"?? :-)

DaRaFF:
> Our SSO guide was had wrong links and was out of date. So I added a new example for AzureAD which is the most common use-case for our customers.

Doesn't use anyone SSO with Github? If we still think it's used we could just update the Github guide too (because it's already there).

content/guides/authentication/azuread-login.md
```diff
@@ -0,0 +1,112 @@
...
+title: Single Sign-On
+description: Learn how to setup SSO with the example of AzureAD.
+weight: 2
+---
+
+## SSO example with Azure AD
+To use a SSO with OpenID Connect we have a strategy `li-authentication-openid-connect` to use in the `auth.connections` config. With this strategy you can use a SSO Service (AzureAD, google, github, facebook, ...) which supports the OpenID Connect for authentication.
```
DaRaFF, line 8:
```suggestion
To use SSO (Single Sign On) with OpenID Connect we have a strategy `li-authentication-openid-connect` to use in the `auth.connections` config. With this strategy, you can use a SSO Service (AzureAD, Google, GitHub, Facebook, ...) which supports OpenID Connect for authentication.  
```

content/guides/authentication/azuread-login.md
```diff
@@ -0,0 +1,112 @@
+---
+title: Single Sign-On
+description: Learn how to setup SSO with the example of AzureAD.
+weight: 2
+---
+
+## SSO example with Azure AD
+To use a SSO with OpenID Connect we have a strategy `li-authentication-openid-connect` to use in the `auth.connections` config. With this strategy you can use a SSO Service (AzureAD, google, github, facebook, ...) which supports the OpenID Connect for authentication.  
+
+This is an example for an AzureAD SSO. It has to be added to the server config in the `auth.connections` config. It creates a setup for the SSO with AzureAD and shows the button for that on the login page. It is also using the existing local user from livingdocs and transfer it to a AzureAD User. This means that the articles which are already written from this user will still be assigned to the same user. It isn't possible to login with the local login after the first login with AzureAD. The match of the user is done with the email address.
```
DaRaFF, line 10: > ...and shows the button for that on the login page.

A screenshot helps a lot for understanding.

DaRaFF, line 10:
```suggestion
To enable AzureAD SSO for Livingdocs, add the config below to the server config in `auth.connections`. It creates a setup for SSO with AzureAD and shows the button for that on the login page. It is also using the existing local user from Livingdocs and transfers it to an AzureAD User. The already written articles with this user will still be assigned to the same user. It isn't possible to log in with the local login after the first login with AzureAD. The match of the user is done with the email address.
```

content/guides/authentication/azuread-login.md
```diff
@@ -0,0 +1,112 @@
...
+
+```js
+const nanoid = require('nanoid/generate')
+const alphabet = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
+const azureConfig = JSON.parse(process.env.auth_azure || '{}')
+
+module.exports = {
```
DaRaFF, line 17: Here is a missing `auth` property, it should be 
```js
module.exports = {
  auth: {
    ...
    connections: { ... }
  }
}

File: content/guides/authentication/azuread-login.md [hunk 5]
```diff
@@ -0,0 +1,112 @@
...
+      // example for google: https://accounts.google.com/.well-known/openid-configuration
+      issuer: 'https://login.microsoftonline.com/{tenant}/v2.0/.well-known/openid-configuration',
+      config: {
+        clientId: azureConfig.clientId,        
+        clientSecret: azureConfig.clientSecret,
+        scope: 'openid email profile',
+        async extractClaims ({tokenSet}) {
+          const claims = tokenSet.claims()
+          // assert group here
+          await transferLocalIdentityToAzure(claims)
```
DaRaFF, line 45: How many times will this function be called, on every login or more?

@marcbachmann this looks quite complicated for me (and is more than a config). Is this currently the best possible solution? Can we improve that in the future?

DaRaFF (approved):
Thanks for this guide!

Feel free to apply my changes, but the wrong config must be fixed.

meinradandermatt:
> > Our SSO guide was had wrong links and was out of date. So I added a new example for AzureAD which is the most common use-case for our customers.
> 
> Doesn't use anyone SSO with GitHub? If we still think it's used we could just update the Github guide too (because it's already there). Or do you think it makes no sense?

From my point of view it will be the same as the AzureAD Guide just a different Issuer. ;-)

content/guides/authentication/azuread-login.md
meinradandermatt: It depends, I provide a file which can be used in the config. So you would use `auth = require.resolve()` so that is why I mentioned it where to put it. But I also see your point. How do we usually do the config examples?

DaRaFF: We usually show the config from the server config root. If you prefer an example with require.resolve, I would make a complete example for clearness.

DaRaFF:
> From my point of view it will be the same as the AzureAD Guide just a different Issuer. ;-)

If it's that simple, you could add this note to the guide.

meinradandermatt:
> > From my point of view it will be the same as the AzureAD Guide just a different Issuer. ;-)
> 
> If it's that simple, you could add this note to the guide.

It is. :-) I will add it.

marcbachmann:
> From my point of view it will be the same as the AzureAD Guide just a different Issuer. ;-)

The github example needs a few tweaks as it only supports oauth2, not openid connect.
```js
      'github-oidcc': {
        strategy: 'li-authentication-openid-connect',
        enabled: false,
        loginEnabled: false,
        registrationEnabled: true,
        autoRegistrationEnabled: false,
        connectionId: 'github-oidcc',
        issuer: {
          authorization_endpoint: 'https://github.com/login/oauth/authorize',
          token_endpoint: 'https://github.com/login/oauth/access_token',
          userinfo_endpoint: 'https://api.github.com/user'
        },
        config: {
          type: 'oauth2',
          scope: 'user:email profile',
          clientId: secrets.githubAuthConfig.clientID,
          clientSecret: secrets.githubAuthConfig.clientSecret,
          async extractClaims ({client, tokenSet}) {
            const userInfo = await client.userinfo(tokenSet)

            // We need to fetch the emails separately as a user can hide it on their
            // profile and in that case it's not present on the userInfo object.
            const emailRes = await axios({
              method: 'get',
              url: 'https://api.github.com/user/emails',
              headers: {
                Accept: 'application/vnd.github.v3+json',
                Authorization: `Bearer ${tokenSet.access_token}`,
                'User-Agent': 'livingdocs-server (github-oidc)'
              }
            })

            const email = emailRes.data.find((e) => e.primary) || {}
            return {
              sub: userInfo.id,
              name: userInfo.name,
              email: email.email,
              email_verified: email.verified,
              picture: userInfo.avatar_url
            }
          },
          // defaultProjectHandle: 'enterprise-web-print',
          // async extractGroupNames (claims) {
          //   return ['owners']
          // }
        },
        ui: {
          label: 'Github',
          icon: 'github'
        }
      },
```

content/guides/authentication/azuread-login.md
marcbachmann: > It isn't possible to log in with the local login after the first login with AzureAD. The match of the user is done with the email address.

Why wouldn't be that possible? We don't disable the local authentication in the code? https://github.com/livingdocsIO/documentation/pull/483/files#diff-cf940d8d1d870d147dde0ee6ad2022dea7efef258d2a1004fab473540b442226R24

meinradandermatt: my fault, I didn't realize that we don't transform it we just add the new identity provider. But we have to rename the function name, as it isn't a transform.

PR #493, opened 2022-05-04
feat(publish control): add release notes and stub guide

benib:


benib:
@DaRaFF @marcbachmann I'd would be great if you could add information about the events and things to setup for the scheduling to the guide.
I will continue to work on it as well with screenshots etc. So let's not force-push here.

PR #494, opened 2022-05-05
Table dashboards: Add documentation for new features and polish 2022-05 release notes

romankaravia:


content/operations/releases/release-2022-05/release-2022-05.md
```diff
@@ -84,33 +84,32 @@ TODO
...
 ### Table Dashboards
 
-TODO@beni: is this description correct?
-
 Table dashboards are a new flexible type of dashboards where individual columns can be configured.
 
-The following functionality has been added:
-* Table dashboard used as default for articles, pages, and data records
+The following functionality has been added in this release:
+* Table dashboard can be used as default for articles, pages, and data records (not enabled by default, to enable, set `documents.useLegacyDashboards` to false in the editor config)
```
benib, line 90: I wouldn't mention this just now. We have some functionality (drag and drop, hugo etc) missing here that some customers rely on. For them, switching doesn't work.

So this has to be either very clear, or we shouldn't mention switching the default now.

PR #503, opened 2022-05-24
Add info about CLI and SDK

looer:
I have used the CLI recently and felt that some info was missing in the docs on how it works (download -> modify -> publish), when to use it and the difference in the component formats.

It's still not clear to me why the HTML should be the recommended format for the CLI, as the components documentation only talks about JS components. If it really makes sense that is the recommended option, then it should be placed as first element in the list to avoid further confusion. @peyerluk do you mind clarifying what was the idea behind?

I also copied some info from the SDK repo readme onto the SDK page, as it was empty. 

Mostly, I wanted to clarify the use cases for the CLI and the SDK. As a new user, my first thought was that the CLI was used to interact with the Public API, but it's not the case.

sdk-docs

### DIFF
--- a/content/reference-docs/cli/_index.md
+++ b/content/reference-docs/cli/_index.md
@@ -11,10 +11,24 @@ menus:
 
 {{< github "livingdocsIO/livingdocs-cli" "Livingdocs command line utility" >}}
 
-## Documentation
+## When to use the CLI
 
-The `livingdocs-cli` is used to update a Livingdocs Project Config. 
+The `livingdocs-cli` is used to update a Livingdocs [Project Config](.././project-config). 
 
 It's the only way of changing project config and components if you are using the [Livingdocs Service](https://edit.livingdocs.io/).
 
-The CLI is not made to interact with the [Public API](.././public-api), for that you can call directly the API or use the [Node SDK](.././sdk).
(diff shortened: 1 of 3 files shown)


content/reference-docs/cli/_index.md
```diff
@@ -12,3 +12,9 @@ menus:
 {{< github "livingdocsIO/livingdocs-cli" "Livingdocs command line utility" >}}
 
 ## Documentation
+
+The CLI is used to modify setting in a live Livingdocs project, without the need to touch its code.
```
DaRaFF, line 16:
```suggestion
The `livingdocs-cli` is used to update a Livingdocs Project Config. 
```

DaRaFF (approved):

DaRaFF:
@looer thank you very much for your docu commit! I'm always happy if engineers feel the need to add/update documentation (that's the pathfinder principle we talked about :-))👍  🎉 

@peyerluk I would like to merge that PR as a first version and then we should have some discussions in the DX Sync
- Do we want to have the SDK in such detail in the reference docs or do we want to completely get rid of the SDK in the doc when we are not maintaining it? I guess that we currently have no real plan for what to do with the SDK.

PR #504, opened 2022-05-31
Update documentPropertyNameConfigs

JSchenk8:
I have updated the config for the liDateTimeRange filter, so wanted to make the documentation slightly clearer. 

Now it states we support updated at and created at, but also explains the default is updated at.

### DIFF
--- a/content/reference-docs/editor-extensions/editor-configuration/display-filter.md
+++ b/content/reference-docs/editor-extensions/editor-configuration/display-filter.md
@@ -11,8 +11,8 @@ Possible values for `displayFilters` are:
   displayFilters: ['liDateTimeRange']
 
   // custom config
-  //   documentPropertyName 'createdAt'/'updatedAt'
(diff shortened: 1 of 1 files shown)


content/reference-docs/editor-extensions/editor-configuration/display-filter.md
```diff
@@ -11,7 +11,7 @@ Possible values for `displayFilters` are:
   displayFilters: ['liDateTimeRange']
 
   // custom config
-  //   documentPropertyName 'createdAt'/'updatedAt'
+  //   documentPropertyName: Supports 'createdAt'/'updatedAt', defaults to updatedAt
   //   metadataPropertyName -> filter by any of your metadata date fields
```
marcbachmann, line 15: Could you use the same structure here. `metadataPropertyName: Supports any of your metadata date fields`

JSchenk8: @marcbachmann 🚀

PR #509, opened 2022-06-21
textcount: add textcount.showEditableCount doc

benib:
This was added a while back, here is the doc: https://github.com/livingdocsIO/livingdocs-editor/pull/4617


DaRaFF (approved):
thx 👍

PR #519, opened 2022-07-12
WIP: fix: Don't use desknet.scheduleEnabled project config property

ajwild:
@DaRaFF I'm leaving this as WIP for now because it will only be in the September release. Do you think I should implement `{{< deprecated-in release-2022-09 >}}`? It might make more sense to keep this config visible until it is no longer used in any active release.

### DIFF
--- a/content/guides/integrations/desknet/index.md
+++ b/content/guides/integrations/desknet/index.md
@@ -68,9 +68,6 @@ Add the following to the project config of the project you want to connect with 
   desknet: {
     enabled: true,
 
-    // Optional. See "Story Planning Schedule in Livingdocs" section below
-    scheduleEnabled: true,
-
     credentials: {
       clientId: '******',
       clientSecret: {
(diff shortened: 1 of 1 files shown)


DaRaFF (approved):

DaRaFF:
> Do you think I should implement {{< deprecated-in release-2022-09 >}}

Difficult to say, currently I would say no. We only reflect the most up-to-date state of our documentation and when we start adding deprecated configs it gets even more complicated than it is already.


We are thinking about different solutions where we want to go with the documentation.
- One option is to provide different versions of the documentation (e.g. for release-2022-07, release-2022-05 and so on).
- Another (must) option is to improve the schema validation and with that, you can give better feedback on what is allowed and what is not.

ajwild:
Ok. Sounds good to me. I'll merge this in mid-September then :+1:

ajwild:
Handled in https://github.com/livingdocsIO/documentation/pull/547

PR #526, opened 2022-07-19
fix(spellcheck): remove unsupported suggest property from spellcheck server example response

benib:


benib:
@meinradandermatt fyi

PR #537, opened 2022-08-10
fix: document teaser guide

looer:
- added the useDashboard config to include service
- fixed code formatting examples
- fixed a typo in a code example (missing comma)
- added an image to show new icon in the toolbar. Maybe would make sense to add more screenshots to understand better each step


JSchenk8:
I think this is good - the key step which was missing was useDashboard - the rest I think is clear from all the other documentation on dashboards and teasers. This was the stumbling block for Dumont, thanks for updating the docs 🚀

JSchenk8 (approved):

PR #549, opened 2022-08-25
Fix formatting of multiline lists

JordiVM:
I saw some broken lists in the release notes and I fixed them in this PR.


DaRaFF (approved):
👍

PR #552, opened 2022-08-29
Add onlyOnComponents to altTextPrefilling

ajwild:
Add new `onlyOnComponents` property to the `altTextPrefilling` config. I've changed the page structure slightly to make it easier to add a more detailed description and the release version which contains the new feature.


ivester (approved):
Nice documentation, makes the other related code reviews easier!

PR #555, opened 2022-08-31
Update `li-transcoding-state` with sample config and release note

dfreier:
As suggested, I added further information to the `li-transcoding-state` plugin docs.


DaRaFF (approved):
thx 👍

PR #569, opened 2022-09-15
fix(mimetypes): Better description

JSchenk8:
Updated the description on how the config for allowed MimeTypes affects the image services and editor


ajwild (commented):
Just below this new sentence there is the text "If neither is defined, all mime types are allowed". Changing this to "If neither is defined, all mime types will be processed by this service" would help to clarify things further. "Processed" could also be "handled", I'm not sure which word is the best option.

JSchenk8:
@ajwild Good point, I've updated it. Went with 'handling' in the end, felt clearer

ajwild (approved):

PR #575, opened 2022-10-07
fix: Add note regarding Retresco compatibility

ajwild:
Relations:
  - Issue:
    - https://livingdocs.slack.com/archives/C01V67WDRC2/p1665149282987469
    - https://livingdocs.slack.com/archives/C01V67WDRC2/p1647351687204249

The incompatibility with Retresco has been brought up a couple of times. I'm not 100% sure if it's the version of the Retresco API that's important, or whether Retresco have to enable a "Livingdocs plugin", so I've deliberately left this vague. My guess is that they need to enable the plugin on their side, and the version isn't important, but it's out of our control anyway.


content/guides/integrations/retresco.md
```diff
@@ -8,6 +8,8 @@ weight: 8
 
 The [Retresco](https://www.retresco.com/) integration uses semantic analysis and keywording to tag documents with relevant entities. These entities are then stored within a document's metadata and can be used to categorise documents.
 
+Please be aware that the Retresco API is also important for the integration to work. The key feature required is the `rtr_payload` object in the Retresco `POST /api/entities` response. This object should exist when there are one or more entities in the `rtr_*` arrays in the response. If the object is not returned then please contact Retresco to upgrade your instance so that is compatible with Livingdocs.
```
looer, line 11: I feel like it's missing something in the first sentence: should it be Retresco API _version_ or Retresco API _response_ ?

looer (changes requested):

content/guides/integrations/retresco.md
ajwild: I did have "version", but then I don't think it's actually the version that's important so I removed it. How about "your Retresco API server setup"?

looer: What if we rewrite the sentence like this:
"Please be aware that Retresco may need to enable an additional feature on your setup for the integration to work. The key feature required is the... "

PR #582, opened 2022-11-01
Add focal point documenation

dfreier:
Focal point documentation, linked in 2022-11 release notes.
Feature PR: https://github.com/livingdocsIO/livingdocs-editor/pull/5916


meinradandermatt (approved):
Can you update the directive structure as well? https://docs.livingdocs.io/reference-docs/document/content/directives-models/#image

As discussed with the named-crops we should find a solution how to name it when we have configured the `image-ratios`

DaRaFF (approved):

PR #592, opened 2022-11-15
Add use cases per `roles` configuration combination

JordiVM:
Extend `Roles` configuration documentation to include use cases per `roles` combination


DaRaFF (approved):
I love it ❤️

PR #593, opened 2022-11-17
Add `all.js` default configurations and comments to the documentation

JordiVM:
Just add `Editor Config` similar to `Server Config` page in the documentation


svenlivingdocs (changes requested):
Thee config looks good.
The moved files however don't show in the "Livingdocs Editor Use Case" navigation item.

svenlivingdocs:
> It is actually broken. It shouldn't show that one. I will fix it.

Ah, thanks!

PR #595, opened 2022-11-30
Fix first publication date

ajwild:
Relations:
  - Issue: https://github.com/livingdocsIO/livingdocs-planning/issues/4767
  - Related PR's:
    - https://github.com/livingdocsIO/livingdocs-server/pull/4957
    - https://github.com/livingdocsIO/livingdocs-server/pull/4992

It would be great if you could both give the text in the latest release notes a review.

Is "Breaking Changes" the correct section? All other manual migrations have been listed here, although it's not technically a breaking change this time.

Do we need to specify pre or post deployment? I suppose post deployment makes the most sense for a fix like this.

### DIFF
--- a/content/operations/releases/master.md
+++ b/content/operations/releases/master.md
@@ -10,7 +10,7 @@ aliases:
 ---
 
 {{< release-header
-  title="January 2022 Release"
+  title="January 2023 Release"
   upcoming=true
   legacy=false
   current=false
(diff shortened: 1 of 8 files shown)


DaRaFF (approved):
> Is "Breaking Changes" the correct section? All other manual migrations have been listed here, although it's not technically a breaking change this time.

Yes because it has the highest attention 

> I suppose post deployment makes the most sense for a fix like this.

Yes, agree.

Can you add that description again to the January Release? The November release is out already, so it's not given that all customers see this cleanup.

PR #599, opened 2022-12-13
A custom table dashboard cell example

JSchenk8:
Here is a simple custom dashboard cell showing scheduled publish date


content/reference-docs/editor-extensions/vue-component-registry.md
```diff
@@ -218,3 +218,82 @@ props: {
   }
 }
 ```
+An example dashboard cell here shows the time property for a scheduled publish document using the new [publish control]({{< ref "/guides/editor/publish-control/publish-control-basics.md" >}})
```
DaRaFF, line 221: Move that guide to the guides and add a link here. The reference docs should not be too long and only explain the extension and its properties. But implementations should be shown via guides.

JSchenk8: Great, thanks! I've updated it, let me know what you think :)

content/guides/editor/custom-table-dashboard-cells.md
```diff
@@ -0,0 +1,98 @@
...
+title: Custom Dashboard Cell
+description: Create a custom table dashboard cell
+weight: 14
+---
+
+{{< added-in release-2022-03 >}}
+
+## Motivation
+
+With the table dashboard introduced in March 2022 we offer custom read-only cells.
```
DaRaFF, line 11: `table dashboard` -> `Table Dashboard`. We decided to write all Livingdocs entities with a big character. 

The motivation for an extension guide should always contain an explanation of when and why one should write an extension. Our default is using the default features with configs, so there should be a good reason to not use that standard.

Long time ago I was writing something similar, maybe you can get some inspiration here -> https://docs.livingdocs.io/guides/search/custom-index/

DaRaFF: We should also always tell the advantages and disadvantages of an extension.

content/guides/editor/custom-table-dashboard-cells.md
```diff
@@ -0,0 +1,98 @@
...
+
+These cells are passed two props, the document and custom options.
+
+## Guide
+
+Below is an example dashboard cell showing the time property for a scheduled publish document using the new [publish control]({{< ref "/guides/editor/publish-control/publish-control-basics.md" >}})
```
DaRaFF, line 17: I still like screenshots where you can see the result (also with a red border to see which cell is custom)

content/guides/editor/custom-table-dashboard-cells.md
```diff
@@ -0,0 +1,98 @@
...
+    const publishDate = this.document.publishControl.publishSchedule.date // UTC String, e.g. 2022-12-14T17:03:00.000Z
+    return format(new Date(publishDate), 'HH:mm')
+    }
+  }
+}
+</script>
+
+```
+
+This must be registered in the vue component registry:
```
DaRaFF, line 57: at the editor

content/guides/editor/custom-table-dashboard-cells.md
```diff
@@ -0,0 +1,98 @@
...
+
+```js
+coreApi.vueComponentRegistry.registerComponent({
+  type: 'tableDashboardCell',
+  name: 'customPublishTimeCell',
+  component: require('../custom-publish-time-cell.vue').default
+})
+```
+
+And can be configured for use in a table dashboard in the `editor_settings` in the server:
```
DaRaFF, line 67: on the server in the Project Config (with a link to the editor settings in the reference docs)

DaRaFF (changes requested):
I know documentation is really hard (btw, you did not do a bad job, but I guess we all need to get better, sorry that I'm using your PR as general criticism).

But see it like that:
Whenever you provide precise documentation, the whole team and customers save a lot of time.
I would say for guides these things are important:
- Motivation when and why you should use that approach and when another approach is better
- All helpful links to other sources (e.g. reference doc)
- Sometimes images to imagine the result (even when it can be outdated, but you get a general idea)

content/guides/editor/custom-table-dashboard-cells.md
```diff
@@ -0,0 +1,98 @@
...
+
+```js
+columns: [
```
DaRaFF, line 70: // projectConfig.editorSettings

JSchenk8:
@DaRaFF  Thank you for the thorough review, I really appreciate improving my documentation. I have followed your advice and hopefully clarified things. Let me know what you think, more feedback welcome

content/guides/editor/custom-table-dashboard-cells.md
```diff
@@ -0,0 +1,117 @@
...
+title: Custom Dashboard Cell
+description: Create a custom Table Dashboard cell
+weight: 14
+---
+
+{{< added-in release-2022-03 >}}
+
+## Motivation
+
+With the Table Dashboard introduced in March 2022 we have created upstream components configurable to suit most customer needs. We would highly recommend using these upstream components.
```
DaRaFF, line 11: Why do we recommend upstream components?
My best guess is:
- less code
- less maintenance
- better integration with other features (indexing, dashboard filter, metadata plugins)

content/guides/editor/custom-table-dashboard-cells.md
```diff
@@ -0,0 +1,117 @@
...
+
+However, we also offer custom read-only cells for times when you have small requirements to display in a Table Dashboard. For example, if you have a custom metadata field you want displayed.
```
DaRaFF, line 13: > if you have a custom metadata field you want displayed

This is already possible. Maybe you should describe in more detail what you mean (e.g. a different presentation).

DaRaFF (approved):
Nice, please consider my comments and after that you can merge it.

PR #608, opened 2023-01-04
Update images for Azure app registration steps

JordiVM:
It was not clear which value to copy in the Secret from the App registration, just updated the images, so it is clear now.


marcbachmann (approved):
should we use a picture that chooses `sAMAccountName` instead of the group id?
feel free to merge already

PR #609, opened 2023-01-13
Remove angular

looer:
A draft of the guide to migrate away from angular, as discussed here:
https://3.basecamp.com/4910529/buckets/25232270/documents/5708393435

I've tried to make an overview of editor customisation that were available with Angular before and now are still supported with core components or Vue components. 

The intent of the guide is not to explain everything from scratch but only to send customers to the right place in the documentation, as they may not be aware of what is possible and where to find it.

Let me know if you see something wrong or missing


content/reference-docs/editor-extensions/removing-angular.md
```diff
@@ -0,0 +1,81 @@
...
+    parent: Editor Extensions
+---
+
+Until now you could have editor extension both in Vue and Angular. Starting from `release-2023-01` we have decided to deprecate the support to Angular extensions to focus on core and Vue components. 
+
+**The support to Angular plugins will be dropped from `release-2023-09`**
+
+Migrating from AngularJS to Vue can be a complex process, especially if you have a lot of downstream custom plugins. 
+
+That's why we want to leave you enough time to make the transition, but it's important to discuss it together beforehand to overcome any blocking points.
```
meinradandermatt, line 16: I would like to make the transition to core plugins. So they don't have the effort on their side it will be in the core. Our goal for next year would be that we don't need any custom vue plugins on editor side anymore.

meinradandermatt: So I would like to switch a bit the order. So focus on core and if really not possible otherwise to vue. :-) You mention that nicely later so I think it is important to have it up here.

looer: Okay, yeah it make sense 👍

meinradandermatt (approved):

content/guides/editor/push-notifications/index.md
```diff
@@ -4,6 +4,10 @@ description: Setup push notifications with a custom dashboard item
 weight: 15
 ---
 
+**Push Notifications are currently deprecated, as they were implemented using Angular. 
+Angular support is going to be dropped in release-2022-09.
```
DaRaFF, line 8: it's 2023-09, right?

content/reference-docs/editor-extensions/removing-angular.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Removing Angular
+description: Support for Angular is going to be dropped, these are some strategies to migrate.
+weight: 5
+menus:
+  reference-docs:
+    parent: Editor Extensions
+---
+
+Until now you could have editor extension both in Vue and Angular. Starting from `release-2023-01` we have decided to deprecate the support to Angular extensions to focus on core and Vue components.
```
DaRaFF, line 10: @looer @peyerluk I know the root of this guide is the removal of Angular. But at least the basic ideas here could also be transported to the introduction section, where you could describe how to configure/extend the system and what approaches we propose.

content/reference-docs/editor-extensions/removing-angular.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Custom Include UI with Vue
+
+In case you need a more complex UI and logic, you can use a Vue component to render the UI. 
+
+[See here]({{< ref "/reference-docs/document/includes/editor-customization#custom-include-user-interface-with-vue" >}}) for an example.
+
+## Dashboard Card and Cells
+
+In `release-2023-05`we introduced [Table Dashboards]({{< ref "/reference-docs/project-config/editor-settings#example-table-dashboard">}}), a more powerful and customisable dashboard. If you have not migrate done it already, you should migrate your dashboard to table dashboards.
```
DaRaFF, line 64: it's 2022-05, right?

DaRaFF: Table Dashboards is a main concept in Livingdocs, therefore we write it in big letters -> "Table Dashboards"

content/reference-docs/editor-extensions/removing-angular.md
```diff
@@ -0,0 +1,84 @@
...
+
+Dashboard Cells can still be customised using Vue components. Custom components
+Custom components of type [tableDashboardCell]({{< ref "reference-docs/editor-extensions/vue-component-registry#tabledashboardcell">}}) can be used to render custom content inside a table cell.
+If some information are missing from your Dashboard Cell, discuss it with your Customer Manager and consider an Implementation Partnership.
+
+## Filters
+
+It is possible to register a custom filter and use it as a [displayFilter]({{< ref "/reference-docs/editor-extensions/editor-configuration/display-filter.md" >}}) for dashboards or search modals.
+
+At the moment there are 4 types of custom filters
```
DaRaFF, line 75: there are only 2 supported types left

PR #614, opened 2023-02-07
Update Metadata Plugins which are supported in Includes

DaRaFF:


content/reference-docs/document/metadata/metadata-plugin-list.md
```diff
@@ -362,15 +362,15 @@ A `li-document-reference` metadata field shows a reference to another document.
 metadata: [
   {
     handle: 'references',
-    type: 'li-document-references',
+    type: 'li-document-reference',
```
benib, line 365: nice one, ideally the line above, the handle would be `reference` as well. And in the example for `li-document-references` it would have the handle `references` instead of `reference`. There was some mixup happening here :-)

benib (approved):
👍

PR #617, opened 2023-02-13
Add Back to Standard Section

DaRaFF:
# Motivation

Our documentation should only contain our newest proposals on how to set up and integrate Livingdocs.
The Back to Standard section is for customers which have Livingdocs for a long time and with that section, we want to show them the path to the current standard.

This is the first version where we can iterate over with time.


meinradandermatt (approved):

peyerluk (approved):
👍

PR #618, opened 2023-02-15
Remove angular from include docs

JSchenk8:
I am working through the bluewin BTS and anytime I find some reference to making things in angular I will remove it. 

This is just a WIP and I will add commits to it as I go then finish the Pr.


DaRaFF (approved):
I would just make a PR every time, it's not expensive and then we have the newest version on master already.

PR #621, opened 2023-02-20
Add editor Wrapper to content type docs

JSchenk8:
A simple comment to let users know they can use an editorWrapper as the base section for a content type.

### DIFF
--- a/content/reference/project-config/content-types.md
+++ b/content/reference/project-config/content-types.md
@@ -233,6 +233,8 @@ Defines a HTML string that wraps the document when used in the editor. Use the c
 
 The example above would put a padding of 20px on documents in the editor to give some space on the sides.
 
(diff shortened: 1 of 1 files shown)


looer (commented):
There is already a section about Editor Wrapper below
https://github.com/livingdocsIO/documentation/pull/621/files#diff-721ed31010e57a8e5fb38d103e8255fd27c73f6f68cf325855e226abc4668fc0R234-R242 

What I would add is an explanation about why one would want/need to use it.

JSchenk8:
@looer Thank you 😂  I didn't check that far down! I've added a sentence about using it to stop content being dropped elsewhere, what do you think?

DaRaFF (approved):

looer (approved):

PR #629, opened 2023-03-06
Update identifier references

peyerluk:
## Changelog

- Use remove `identifier` in components where outdated.
- Remove `transmogrifier.wrap()` from documentation


content/guides/integrations/dpa-import.md
```diff
@@ -110,20 +110,6 @@ mapDocument: function (abstractArticle) {
         delete component.content.source
         return component
       })
-      .wrap({
```
peyerluk, line 113: @gabrielhase I couldn't find this in the framework. Did I overlook something?

PR #642, opened 2023-04-27
Metadata previews

benib:
https://github.com/livingdocsIO/livingdocs-editor/pull/6738

### DIFF
--- a/content/reference/project-config/content-types.md
+++ b/content/reference/project-config/content-types.md
@@ -65,6 +65,21 @@ contentTypes: [
       }
     ],
 
+    // added in release-2023-05
+    metadataPreviews: [
+      {
+        iframe: {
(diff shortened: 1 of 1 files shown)


content/reference/project-config/content-types.md
```diff
@@ -288,6 +303,61 @@ metadataGroups: [
 ]
 ```
 
+## Metadata Previews
+
+{{< added-in release-2023-05 >}}
+With Metadata Previews, you can visible show the user how the value of a certain metadata property will be visually represented.
```
marcbachmann, line 309:
```suggestion
With Metadata Previews, you can show the user how the value of a certain metadata property will be visually represented.
```

marcbachmann (approved):
Good after changing the units

PR #667, opened 2023-05-18
fix(includes): moved paramschema to be more prominent

JSchenk8:
After writing some paramsSchema includes as part of Bluewin BTS and infobox improvements, along with supporting Netcetera writing ones with li-document-reference/s, I wanted to just tidy up the documentation a little bit. 

I moved the paramsSchema to the top to try and promote it as the most viable option. And reminded users not to write any custom code without talking to us first. 

I also added the option of rendering a publication in the renderFunction. 

Happy to discuss 👍

### DIFF
--- a/content/reference/document/includes/server-customization.md
+++ b/content/reference/document/includes/server-customization.md
@@ -9,7 +9,8 @@ menus:
 This section gives an exhaustive listing of all options for Includes on the server side.
 To get more information about Includes in general, visit the [Includes Overview]({{< ref "/reference/document/includes" >}}).
 
-## Example
+We recommend where possible going down the paramsSchema route for includes. This saves writing custom sidebar components and covers most  include cases. Where it does not, please get in touch with your customer solutions manager before writing custom components to see if we can extend this option for your use case.
+## Registering your include
 
 ```js
 // runtime config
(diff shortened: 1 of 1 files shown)


peyerluk (approved):

content/reference/document/includes/server-customization.md
```diff
@@ -9,7 +9,8 @@ menus:
 This section gives an exhaustive listing of all options for Includes on the server side.
 To get more information about Includes in general, visit the [Includes Overview]({{< ref "/reference/document/includes" >}}).
 
-## Example
+We recommend where possible going down the paramsSchema route for includes. This saves writing custom sidebar components and covers most  include cases. Where it does not, please get in touch with your customer solutions manager before writing custom components to see if we can extend this option for your use case.
```
peyerluk, line 12: > if we can extend this option

-> if we can extend the paramsSchema for your use case.

peyerluk, line 12:
```suggestion
We recommend where possible going down the paramsSchema route for includes. This saves writing custom sidebar components and covers most  include cases. Where it does not, please get in touch with your customer solutions manager before writing custom components to see if we can extend the paramsSchema for your use case.
```

content/reference/document/includes/server-customization.md
```diff
@@ -22,6 +23,82 @@ liServer.features.register('include-services', async function (feature, server)
...
+    }
+  ],
+  rendering: {
+    type: 'function',
+    async render (params, context) {
+      // Here you can use the publicationApi to get your document
+      // You can then return the content of the linked document with ease, Livingdocs will render
+      // it exactly as you see it in the document itself
+      const id = params.infobox.reference.id
+      const publication = await publicationApi.getPublicationsByDocumentIds([id])
```
DaRaFF, line 57: @JSchenk8 There is a better option to load and access publications:

```js
const documentId = params.infobox.reference.id
const documentVersion = publicationApi.getLatestPublication({documentId})
const content = documentVersion.content
```

And please consider these things:
- When fetching documents/publications, always use the name `documentVersion`. With that, customers have the right model in mind when reading code (so as we have).
- Never access internal props (`publication[0].revisionEntity`). The [DocumentVersion](https://docs.livingdocs.io/customising/server/document-version/) model has a lot of useful helpers.

PR #668, opened 2023-05-18
fix(server hooks): preunpublishhook added to table

JSchenk8:
Added the preUnpublishHookAsync to the table of publish hooks. 

The rest of the documentation is pretty good IMO, but I don't know if we can make it possible to search for hooks in the documentation search bar. As it is, 'preparepublishhook' or any other function name don't return anything. Any ideas @DaRaFF ?


DaRaFF (approved):
I found a solution :-)  -> https://livingdocs.slack.com/archives/C4TGTEC3F/p1684764722109699

PR #675, opened 2023-06-12
Add menu tool guide

dfreier:


content/guides/editor/menu-tool/index.md
```diff
@@ -0,0 +1,138 @@
...
+weight: 4
+---
+
+The idea is to use the [`li-tree`]({{< ref "/reference/document/metadata/metadata-plugin-list#li-tree">}}) metadata plugin on a Data Record to define a menu tree structure along with additional metadata.
+Since each menu is just a document, it can be shown in dashboards and referenced in other documents.
+As with all documents in Livingdocs, this is a way to express what the content is, but not how it is rendered.
+
+Possible use cases can be:
+- Top-level navigations in delivery frontends
+- Navigation menus within different topic pages
```
benib, line 13: In this case, the `li-tree` would probably be configured on the page itself, instead of being a separate data record that is then referenced.

benib (approved):
👍 Could it be that the data-record used for the screenshots has no publish control configuration, therefore it shows the old prepare publish flow style UI?

If so, I would update this.
Otherwise, this is cool. Could be extended with some information about how to access this Menu via Public API with the new search query parameters. @ajwild knows the details there probably.

content/guides/editor/menu-tool/index.md
dfreier: What I had in mind here would be something like a menu include service that can reference a menu document and can be placed in the editor and shared across pages.
But I can also remove this use cases section if it produces more questions than answers 😄

dfreier:
> 👍 Could it be that the data-record used for the screenshots has no publish control configuration, therefore it shows the old prepare publish flow style UI?

Yes, gonna update it.

> If so, I would update this. Otherwise, this is cool. Could be extended with some information about how to access this Menu via Public API with the new search query parameters. 

I'll check with Alex 👍

ajwild:
Does this need an `added-in release-2023-07` at the top of the guide, or was it possible before then (except the Public API part)?

dfreier:
Everything was also possible before 👍

PR #676, opened 2023-06-12
Add li-tree multilang docs

dfreier:


benib (approved):
If I see this correctly, the `multilang` config option to turn the multilang on is not mentioned. My understanding is, that this is needed (an rightly so) to turn on multilanguage for li-tree.
Otherwise, this looks perfect 👍

dfreier:
I added the option in the sample config above, but I could loose a few words on that in the dedicated section too 👍

PR #690, opened 2023-06-30
fix(push messages): li string list added

JSchenk8:
Added the compatibility for li-string-list to li-push-messages paramsSchema


dfreier (approved):
❤️

PR #694, opened 2023-07-03
Metadata Refactor

ajwild:
The Metadata Plugin List has been changed to a dynamically rendered section template based on individual plugin files which contain specific front matter (metadata) that describes each plugin. The path has been updated from `/reference/document/metadata/metadata-plugin-list` to `/reference/document/metadata/plugins`, with all links updated to point to the new location, and a redirect added to forward anyone with the old link. The new directory structure allows us to have paths to plugins such as `/reference/document/metadata/plugins/li-tree`.

The individual metadata files aren't perfect yet, but we can iterate on the files now that they exist:
- Write any missing summaries and descriptions
- Replace links to GitHub images
- Add more details about indexing 
- Add more details about searching
- Maybe split the "Default UI" section into multiple sections for the metadata form, dashboard cells, side panels etc.

I shared some template code using themes/hugo-docs/layouts/partials/default-above-content.html and themes/hugo-docs/layouts/partials/default-below-content.html. It might be possible to use [Base Templates and Blocks](https://gohugo.io/templates/base/) to do this properly, but I've had enough of Hugo for now!


content/reference/document/metadata/plugins/li-boolean.md
```diff
@@ -0,0 +1,37 @@
...
+  reference:
+    parent: Metadata Plugins
+summary: A simple Boolean value store, represented in the UI with a checkbox which you can toggle on/off.
+support:
+  document: true
+  media: true
+  tableDashboard: true
+  include: true
+  displayFilter: false
+  searchFilter: true
```
DaRaFF, line 14: When `searchFilter` or `searchSort` is supported/true, what does that exactly mean?
Do you think it makes sense to add some more info on what a supported function means?

content/reference/document/metadata/plugins/li-color.md
```diff
@@ -0,0 +1,46 @@
+---
+title: li-color
+type: metadata-plugins
+menus:
+  reference:
+    parent: Metadata Plugins
+summary: Single category selection.
```
DaRaFF, line 7: thats obviously wrong :-)

DaRaFF (approved):
I like the new structure and approach and we can go for it 👍 

I didn't go into detail and controlled every plugin if the content is correct. I assume your goal is to provide the new structure/setup and the fine detail for every plugin can be done later. If not, you have to make another request to me.

I see 2 disadvantages of that approach (which is fine for me at the moment)
- flexibility of content (but we want to have standardized plugin descriptions anyway)
- comprehensibility of code (a core dev now needs to understand what is possible based on the existing plugins or the hugo templates)

content/reference/document/metadata/plugins/li-boolean.md
benib: Not 100% sure about the wording here. Would it make sense to go with `baseFilter` instead? Or `dynamic indexing`?
In other contexts, we keep explaining the difference between displayFilter and `baseFilter`, so I believe we should either push very hard to rename `baseFilter` to `searchFilter` or explain the difference very good (are we using `searchFilter` on the public API search endpoints?).

ajwild: I've updated `searchFilter` to a `dynamicIndexing` property, and I've removed `searchSort`. We can easily link to some additional documentation from the template if needed. I'll get this merged and then ask Sven to take a look, as there's probably something better we can be doing at the top of each document than a simple table.

PR #702, opened 2023-07-06
Tidy up the use of release shortcodes

ajwild:
I don't think we need to tell people that certain functionality was added in 2021 any more. I propose that we only include `added-in` for the previous three releases, so now it would be March, May, and July. This allows customers the time to upgrade to one of the two supported releases. In two months we can go through and delete the `added-in` for March  and so on.

The `removed-in` shortcode should be deleted with the same timeline, but `deprecated-in` will likely hang around much longer until we actually make the breaking change. I think there may be numerous instances where we say something is deprecated in the documentation, but it was never communicated to customers through the release notes, so we need to be careful about this.

I have also tried to make sure that wherever we mention a release we use the relevant shortcode so that it's easier to maintain this going forward.


DaRaFF (approved):
nice, thx 👍

PR #710, opened 2023-07-14
Improve metadata search

ajwild:
- Added `npm` scripts for quicker production testing
- Fixed `keywords` not used for page search results (they were only used sections)
- Added 'Metadata Plugin' as a keyword for all metadata plugins for a little boost
- Moved section titles to `subtitle` property for customisable boost value
- Renamed 'Metadata Plugins' back to 'Metadata Plugins List'
- Applied boost values to fields during indexing to help simplify the search code
- Use the highest score from a page instead of adding sections together
- Fix link styles on Metadata Plugins List page

### DIFF
--- a/bin/build-index.js
+++ b/bin/build-index.js
@@ -1,7 +1,7 @@
 const fs = require('fs')
 const {Parser} = require("htmlparser2")
 const {DomHandler} = require("domhandler")
-const {getText, hasAttrib, getAttributeValue} = require('domutils')
+const {getText, hasAttrib, getAttributeValue, textContent} = require('domutils')
 
 buildIndex('./public/search.json')
 
@@ -14,37 +14,18 @@ function buildIndex (file) {
   fs.writeFileSync(file, JSON.stringify(index))
 }
 
-function parseDocument (index, {url, section, categories, title, description, body, keywords}) {
-  const dom = getDom(body)
-
-  let current = {
-    url,
-    section,
-    categories,
-    title,
-    description: description || '',
-    body: '',
-    keywords: ''
-  }
-
+function parseDocument (index, data) {
+  const dom = getDom(data.body)
+  let current = prepareDataForIndex(data)
   index.push(current)
+
   for (const elem of dom) {
     if (elem.type === 'text') {
       if (!elem.data.trim()) continue
       current.body += `\n${elem.data.trim()}`
     } else if (elem.type === 'tag') {
       if (elem.name.startsWith('h')) {
-        const sectionTitle = getText(elem).trim()
-        current = {
-          url: hasAttrib(elem, 'id') ? `${url}#${getAttributeValue(elem, 'id')}` : url,
-          section: section,
-          categories,
-          title: sectionTitle,
-          description: sectionTitle === title ? (description || '').trim() : '',
-          body: '',
-          keywords: (keywords || []).join(', ')
-        }
-
+        current = prepareDataForIndex(data, elem)
         index.push(current)
       } else {
         current.body += `\n${getText(elem)}`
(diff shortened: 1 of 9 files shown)


bin/build-index.js
```diff
@@ -14,45 +14,54 @@ function buildIndex (file) {
...
-  console.log('Processed document %s: %s', title, url)
+  console.log('Processed document %s: %s', data.title, data.url)
+}
+
+function prepareDataForIndex (data, elem) {
+  // Keyword customisations
+  const additionalKeywords = []
+  if (data.type === 'metadata-plugins') {
+    additionalKeywords.push('Metadata Plugin')
+  }
```
ajwild, line 44: We could add this string to all the metadata plugin documents if we want it to be more transparent.

package.json
```diff
@@ -10,11 +10,14 @@
     "glob": "^10.0.0",
     "htmlparser2": "^9.0.0",
     "lunr": "^2.3.9",
-    "opener": "^1.5.2"
+    "opener": "^1.5.2",
+    "serve": "^14.2.0"
```
ajwild, line 14: We don't need to add the dependency, we could use `npx` instead, or we can use another dependency if people have a preference.

themes/hugo-docs/assets/search.js
```diff
@@ -66,10 +64,12 @@ function searchInIndex (index, {query, filterTags, limit}) {
...
         score: 0,
         results: [],
-        title: `${tags}${doc.title === ref.title ? '' : `<a href="${documentUrl}" class="search-results-document__title">${doc.title}</a>`}`
+        title: `${tags}${doc.title === (titleOrSubtitle) ? '' : `<a href="${documentUrl}" class="search-results-document__title">${doc.title}</a>`}`
       }
       byDoc[documentUrl].results.push({url, title, description})
-      byDoc[documentUrl].score += match.score
+			if (match.score > byDoc[documentUrl].score) {
+				byDoc[documentUrl].score = match.score
+			}
```
ajwild, line 72: This change made a massive difference. Otherwise large pages always ended up at the top of results, above exact matches for a title.

ajwild (commented):
I tried a lot of this, including editing the stopwords list, for example excluding 'li' from the index because it appears everywhere. I'm not sure it makes a big difference, but I can share the code as it's only 2 lines. I also tried replacing the built-in tokenizer so that it would also do exact matches for hyphenated values like 'li-text', but the results weren't massively different. In the end I feel the results are much better now without too much manipulation of internal lunr functions.

Feel free to try `npm run build:fast` and then `npm run start:production` to compare the search results to production.

themes/hugo-docs/assets/search.js
marcbachmann: Yes, this most likely broke the bm25 score. To fix the scores a bit, we could index all the text separate and the headings separately with priorization. 

But to link to specific sections, we'd need a secondary index with two sequential searches.

I'm thinking about doing something similar in our elasticsearch index:
```js
{
  url: '/some-page',
  h1: ['Some Page Title'],
  h2: ['Some subtitle', 'Second subtitle'],
  h3: ['Another subtitle'],
  text: ['Paragraph 1', 'Paragraph 2']
}
```

marcbachmann: But then it might not find sections as well as we do it now.

marcbachmann (approved):

marcbachmann:
just saw that `registerInitializedHook` doesn't result in a search match. But this was already broken before. Might need some investigation whether we should apply some normalization before indexing. 🤔

can't find it in the search.json despite that it's a title 🤔 
we should at least add it to the section text.

ajwild:
> just saw that `registerInitializedHook` doesn't result in a search match

That's because `.` is not a word separator in lunr, only space or hyphen. It's easy to change the regular expression. If you search for "liServer.registerInitializedHook" you'll find it.

PR #712, opened 2023-07-17
Add mediaLibrary.image.processing.failOn to server config

ajwild:
Relations:
 - Related PR: https://github.com/livingdocsIO/livingdocs-server/pull/5909


marcbachmann:
just saw that `registerInitializedHook` doesn't result in a search match. But this was already broken before. Might need some investigation whether we should apply some normalization before indexing. 🤔

marcbachmann:
can't find it in the search.json despite that it's a title 🤔

PR #713, opened 2023-07-17
Separate search strings on any non-word character

ajwild:
Relations:
- Related PR: https://github.com/livingdocsIO/documentation/pull/710

This allows matches for "registerInitializedHook", when previously you would need to search for "liServer.registerInitializedHook".

This needs some testing before merging.


marcbachmann (approved):

ajwild:
The hyphen was always in the separator regular expression, so "release-2023-07" would have been split into three tokens anyway. I'll merge and we can see if we notice any strange behaviour in the real world!

PR #715, opened 2023-07-20
Metadata display filters

benib:
Nice system with this metadata table.
I believe showing support for display filters would be cool? Or is there a specific reason not to?


ajwild:
I had added a Display Filters column originally, along with Dynamic Indexing, but I discussed with someone and the decision was that there were too many columns and it was overwhelming. So we cut back to the core columns. I have no strong feelings either way. I guess we need to make a decision on how many columns is too many columns.

marcbachmann:
We've added this in https://github.com/livingdocsIO/documentation/pull/737

PR #716, opened 2023-07-27
Add more breaking changes to Angular Migration plan

ajwild:
I found these two in the console logs, so I figured they should be added to the documentation.


benib (approved):

content/customising/angular-migration/_index.md
```diff
@@ -40,5 +40,10 @@ Migrating away from AngularJS can be a complex process, especially if you have a
 ### release-2023-09
 
 🔥 Remove support for Angular Metadata
+🔥 Remove support for Legacy Dashboards
+
+### release-2023-11
+
+🔥 Remove support for Prepare Publish Flow
```
ajwild, line 47: Thinking about it, does Prepare Publish Flow actually have anything to do with AngularJS, or should it just be the legacy dashboards? The legacy dashboards use AngularJS components, right?

DaRaFF: For me, it's fine to mention the Prepare Publish Flow separately. It's still Angular, but the technology doesn't matter. The important thing here is that the feature will not be supported after November.

DaRaFF (approved):

content/customising/angular-migration/_index.md
DaRaFF: If you're strict, you would be not allowed to mention it here, because it has nothing todo with Angular, but again, it's ok for me.

benib: Let's not mention it. You both are right, it has nothing to do specifically with Angular. We should not have too many deprecation lists to maintain.

PR #718, opened 2023-07-27
Add more details about migrating filters for references

ajwild:
This has come up a couple of times, and even I struggle to remember to query the nested property value, and I wrote the documentation and wrote some of the code! :grin: Hopefully having a section title makes it much more obvious, and gives people a link on the right when there are issues migrating.


DaRaFF (approved):
👍 Thanks for the clarification.

The filter topic is quite hard to grasp and can always be improved. I think we have to work more on that and all input is welcome.

PR #740, opened 2023-09-06
Rename dynamic indexing to search indexing

marcbachmann:


themes/hugo-docs/layouts/metadata-plugins/single.html
```diff
@@ -38,7 +38,7 @@
             <td style="text-align: center;">{{ partial "check-cross" .Params.support.displayFilter }}</td>
           </tr>
           <tr>
-            <td>Dynamic Indexing</td>
+            <td>Search Indexing</td>
             <td style="text-align: center;">{{ partial "check-cross" .Params.support.dynamicIndexing }}</td>
```
JordiVM, line 42:
```suggestion
            <td style="text-align: center;">{{ partial "check-cross" .Params.support.searchIndexing }}</td>
```

JordiVM: Same issue in `themes/hugo-docs/layouts/metadata-plugins/section.html` line 29. I can't add a comment there

PR #742, opened 2023-09-14
Configuring metadata display filters for multiple content types

ajwild:
Note added as a result of a conversation in Slack: https://livingdocs.slack.com/archives/C4TGTEC3F/p1694691100998669?thread_ts=1694614903.890039&cid=C4TGTEC3F


benib (commented):
This reads to me like: "Add the metadata property to all the contentTypes shown" is the solution.

Would it be more correct to say: "Use separate dashboards and with this separate sets of filters for contentTypes that are similar enough so the key metadata properties are the same for all of them"?

ajwild:
I've had another go :+1:

PR #747, opened 2023-09-26
Add retresco re-enrich endpoint to the public API documentation

JordiVM:
Include re-enrich endpoint to Public API documentation.

### DIFF
--- a/content/reference/public-api/retresco.md
+++ b/content/reference/public-api/retresco.md
@@ -1,0 +1,78 @@
+---
+title: Retresco re-enrich
+weight: 15
+renderTOC: false
+menus:
+  reference:
+    parent: Public API
+---
+
+{{< api-example
+  title="Re-enrich documents with Retresco"
+  scopes="retresco"
+>}}
+
+{{< added-in "release-2023-03" >}}
+
+--query--
+```bash
+ACCESS_TOKEN=ey1234
+curl -k -X POST "https://edit.livingdocs.io/proxy/api/api/v1/retresco/re-enrich" \
+  -H "Accept: application/json" \
+  -H "Authorization: Bearer $ACCESS_TOKEN" \
+  -H 'Content-Type: application/json; charset=utf-8' \
+  --data-binary @- << EOF
+  {
+    "doc_ids": ["1", "2", "3"]
+  }
+EOF
+```
+
+--endpoint--
+```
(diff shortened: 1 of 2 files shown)


content/reference/public-api/retresco.md
```diff
@@ -0,0 +1,74 @@
...
+--query--
+```bash
+ACCESS_TOKEN=ey1234
+curl -k -X POST "https://edit.livingdocs.io/proxy/api/api/v1/retresco/re-enrich" \
+  -H "Accept: application/json" \
+  -H "Authorization: Bearer $ACCESS_TOKEN" \
+  -H 'Content-Type: application/json; charset=utf-8' \
+  --data-binary @- << EOF
+  {
+    "doc_ids": ["1", "2", "3"]
```
DaRaFF, line 26: is there a limit how many documentIds you can pass?

content/reference/public-api/retresco.md
```diff
@@ -0,0 +1,74 @@
...
+|Name|Type|Notes|
+|-|-|-|
+|doc_ids|array\<string\|number\>|A list of the desired document ids to re-enrich.|
+
+--description--
+
+Re-enriches documents with Retresco. This endpoint is only available for customers with a Retresco integration enabled. The endpoint will return a 200 status code when the request is valid, then the Livingdocs Server will start the re-enrichment process in the background. The re-enrichment process can take up to 1s per document to re-enrich. If the provided document ids are not found, the re-enrich job will not be executed for that document.
+
+Follow the [guide to configure the re-enrich webhook]({{< ref "/guides/integrations/retresco#re-enrich-documents" >}}) in Retresco's website.
+
```
DaRaFF, line 47: you could add a use case section, like I did here -> https://docs.livingdocs.io/reference/public-api/publications/latest-publications/. I have planned to document a use case for all endpoints to get more clearity why an endpoint should be used. I know in that case it's a little bit of duplication from the description, but I guess it's still worth.

DaRaFF (approved):
Thanks for adding that @JordiVM 👍

content/reference/public-api/retresco.md
JordiVM: Not really, it will process documents `integrations:retresco:enrichConcurrency = 5` at a time and it will finish whenever (max recommended concurrency 10). We don't really enforce a limit here.

PR #749, opened 2023-09-27
Add Command API documentation page

dfreier:


content/reference/public-api/command-api.md
```diff
@@ -0,0 +1,146 @@
...
+```
+PATCH api/v1/documents/:id/commands
+```
+
+--parameters--
+|Name|Type|Required|Notes|
+|-|-|-|-|
+|version|string||Current document version. When set on update the version is checked.|
+|preconditions|array||An array of preconditions for command execution. Each entry is an object with at least a **type** property. Possible types: `isPublished`. See further details in example requests.|
+|commands|array|x|An array of commands to execute. Each entry is an object with at least an **operation** property. Possible operations: `setMetadataProperty`, `setEditableDirective` or `publish`. See further details in example requests.|
```
ajwild, line 34: Also `'setTitle'`

ajwild (approved):
Thanks! Please could you also update https://github.com/livingdocsIO/openapi/blob/master/livingdocs-openapi.json :pray:

content/reference/public-api/command-api.md
dfreier: At the moment `setTitle` is not a valid operation in the json schema for the public API endpoint. Not sure if we should add it since it would not have an effect when using display title pattern.

PR #754, opened 2023-10-06
Add removed-in tag to `li-reference` and `li-reference-list` metadata plugins

JordiVM:
Relations:
- Related PRs: https://github.com/livingdocsIO/livingdocs-server/pull/6186

### DIFF
--- a/content/reference/document/metadata/plugins/li-reference-list.md
+++ b/content/reference/document/metadata/plugins/li-reference-list.md
@@ -5,7 +5,7 @@ menus:
   reference:
     parent: Metadata Plugins List
 summary: |
-  {{< deprecated-in "release-2023-09" >}}. Use [`li-document-references`]({{< ref "/reference/document/metadata/plugins/li-document-references" >}}) instead.
+  {{< removed-in "release-2023-11" >}}. Use [`li-document-references`]({{< ref "/reference/document/metadata/plugins/li-document-references" >}}) instead.
 support:
   document: true
   media: true
@@ -16,6 +16,7 @@ support:
   systemMetadata: false
   planningSystem: false
 deprecatedIn: release-2023-09
+removedIn: release-2023-11
 description: |
   With {{< release "release-2022-11" >}} this is superseeded by [`li-document-references`]({{< ref "/reference/document/metadata/plugins/li-document-references" >}}).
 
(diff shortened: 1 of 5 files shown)


ajwild (approved):

JordiVM:
I will include the Breaking change message into the Release cut PR, using editor's and server's commit message.

JordiVM:
I wasn't sure how to do it. It made sense to me to just have the breaking change, since the deprecation doesn't imply code change. But if we have done it with both on other places, I can add it here too :)

content/reference/document/metadata/plugins/li-reference-list.md
```diff
@@ -6,6 +6,7 @@ menus:
     parent: Metadata Plugins List
 summary: |
   {{< deprecated-in "release-2023-09" >}}. Use [`li-document-references`]({{< ref "/reference/document/metadata/plugins/li-document-references" >}}) instead.
+  {{< removed-in "release-2023-11" >}}. Use [`li-document-references`]({{< ref "/reference/document/metadata/plugins/li-document-references" >}}) instead.
```
ajwild, line 9: Can you check if this looks ok.
```suggestion
  {{< deprecated-in "release-2023-09" block >}}
  {{< removed-in "release-2023-11" block >}}
  Use [`li-document-references`]({{< ref "/reference/document/metadata/plugins/li-document-references" >}}) instead.
```
I don't think we need to repeat "Use li-document-references instead".

content/reference/document/metadata/plugins/li-reference.md
```diff
@@ -6,6 +6,7 @@ menus:
     parent: Metadata Plugins List
 summary: |
   {{< deprecated-in "release-2023-09" >}}. Use [`li-document-reference`]({{< ref "/reference/document/metadata/plugins/li-document-reference" >}}) instead.
+  {{< removed-in "release-2023-11" >}}. Use [`li-document-reference`]({{< ref "/reference/document/metadata/plugins/li-document-reference" >}}) instead.
```
ajwild, line 9: Same as above.

ajwild:
Sorry, I didn't really think about where these deprecation and removed messages were being displayed. I think you were right the first time. Lets only have the removed in shortcodes in this summary, otherwise it looks a bit busy. Sorry for causing extra work :grimacing:

ajwild (approved):

PR #763, opened 2023-11-02
Technical release notes improvements / Feature info shortcode

peyerluk:
# Changelog

- 🎁 Add `feature-info` shortcode
- Use feature-info shortcode for technical release notes in November release

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,10 @@ Check your titles by reading the table of contents generated on the right side o
 {{< release "release-2021-03" >}}
 ```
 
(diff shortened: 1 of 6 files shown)


DaRaFF:
I like that! As a next generation idea: Be able to set links in feature-info to the documentation already (e.g. Public API, Document Version, Server Config and so on). Maybe we can also provide a fixed set of feature-info which is doing the linking automatically and you only have to write `{{< feature-info "Document Version" >}}`

PR #768, opened 2023-11-03
Use new liServer.register methods

peyerluk:
# Changelog

- 🎁 Use new `liServer.register...()` methods everywhere.
- 🎁 Replace `liServer.features.register()` with `liServer.registerServerRoutes()`
- 🔧 Use `daily-planet` as example project handle

PR #758, opened 2023-10-31
Add API changes to November release notes

JordiVM:
Title is the description.

PR #769, opened 2023-11-08
Add original api functions replaced by liServer extensions

JordiVM:
As Ralph mentioned in PR [#768](https://github.com/livingdocsIO/documentation/pull/758), it makes sense to include the original functions replaced by liServer extensions


DaRaFF (approved):
Ah cool, like it that much 👍

PR #773, opened 2023-11-13
Update release-2023-11 notes

dfreier:


JordiVM (approved):
Yep, this makes more sense. Ship it

PR #799, opened 2024-02-23
Use new filter dsl in examples

dfreier:
Updated some config examples where the old filter syntax was still used.


DaRaFF (approved):
Thanks a lot 👍

PR #811, opened 2024-03-06
fix(image url building in deliveries): added

JSchenk8:
A guide to building the URL in deliveries after our changes in the March release. 

I also added a link in the march release.


DaRaFF (approved):
Nice! I think in the future we have to overhaul the guide to make it more clear how the old and the new world looks like. Now t's a little bit a mix of both and difficult to understand.

PR #813, opened 2024-03-11
docs: adds example for migration test

wpp:
We recently received a heads-up from Joe about writing tests for migrations. He provided an example, and I thought it might be helpful for others too.

As always, feedback is welcome.


JSchenk8 (approved):
Sorry, I totally forgot about this! Thank you very much!

PR #814, opened 2024-03-12
docs(defaultComponents): remove audio and html component types

dfreier:
I don't think we support audio and html component types here, at least I have not found any other references in the documentation and at a first glance also no explict handling in the code


JordiVM:
It looks like the schema supports it: https://github.com/livingdocsIO/livingdocs-server/blob/609ae837a7fd4acb643bcfad7bbf410eda063b02/app/features/channel-configs/config-properties/content-types.js#L42-L48

JordiVM:
We can merge this one and have some efforts to fix this in the future maybe? :)

PR #819, opened 2024-03-21
Add li-document-search plugin

robinbisping:
Relations:

- Issue: https://3.basecamp.com/4910529/buckets/33963114/documents/7057829601
- Server PR: https://github.com/livingdocsIO/livingdocs-server/pull/6715
- Editor PR: https://github.com/livingdocsIO/livingdocs-editor/pull/8228


peyerluk:
@robinbisping Any progress here? Would be great if we can merge this.

robinbisping:
@peyerluk Sorry, it got a bit stuck. I'll try to get a review and then we can merge.

peyerluk:
@robinbisping Is li-document-search always preloaded? For other metadata plugins are `preload: true` is required for preloading. So this might be a bit confusing. At least it was to me.

peyerluk:
...adding to the previous comment. This behavior should be added to the documentation as well (if I understand it correctly).

peyerluk:
I think the info that this always searches against published documents is missing in the  docs. Our other references offer the possibility to decide if drafts or publications are linked. So at least it's worth mentioning to avoid confusion.

peyerluk:
One more question :)

The baseFilter and storageFilter formats are only described as Arrays. We should try to have a documentation of these formats in detail. Does this already exist somewhere in the docs? Otherwise maybe we can try to add this?
Are these formats compatible with the search?  E.g. Can I use the stored info to make a request against our search?

robinbisping:
Thanks for the feedback! I'll clarify these points in the docs:

- Documents are always preloaded for li-document-search
- It searches only against the publication index
- You can use the stored data for your own search queries. It uses our Search DSL, which is also documented, but for the displayFilterStates its a bit more complex, as also other properties are supported (sort, context). When you implement your own queries, it's probably best to first consult our [implementation](https://github.com/livingdocsIO/livingdocs-server/blob/60fb40d8c432e7460160c6408046d45f0fd05b6a/app/features/includes/lib/resolve-includes.js#L131-L157) before doing so. I'll try to clarify all this in the docs.

robinbisping:
Display filter states are unfortunetly nowhere documented. I'll add some documentation next week but would like to add it in another PR to have this one merged before.

content/reference/document/metadata/plugins/li-document-search.md
```diff
@@ -0,0 +1,95 @@
...
+
+![li-document-search UI](/reference/document/metadata/li-document-search-example.png)
+
+{{< info >}}
+When utilizing custom display filters, it is crucial to conduct thorough testing. Your display filter must be capable of restoring its state solely from the `filter`, `sort` and `context` attributes. All other attributes are not persisted. For more details, refer to our guide on how to write [custom display filters]({{< ref "/guides/editor/custom-dashboard-filters#custom-vue-component-filter" >}}).
+{{</ info >}}
+
+## Resolving
+
+The specified conditions are stored within documents and evaluated on request. Therefore, teaser lists may contain other teasers over time as new articles match the conditions, without requiring editors to republish a page. Thus, it is recommended to configure your delivery to repeatedly invalidate the cache for such pages and refetch them occasionally from the Composition API to obtain the latest state.
```
marcbachmann, line 91: I don't like the word condition in that context. As we've just introduced the conditional components, this might be misleading. Please just use `filter` instead.

content/reference/document/metadata/plugins/li-document-search.md
```diff
@@ -0,0 +1,95 @@
...
+    parent: Metadata Plugins List
+summary: Dynamic teaser lists based on configurable conditions.
+support:
+  include: true
+description: |
+  The `li-document-search` plugin simplifies the management and rendering of article teaser lists within pages based on configurable conditions. It provides an efficient solution for semi-automatic page management. This approach reduces the need for custom code compared to other solutions within Livingdocs.
+storageFormat: |
+  {
+    baseFilters: <Array>,
+    displayFilterStates: <Array>,
```
marcbachmann, line 15:
```suggestion
    displayFilterStates: <Array<{filter: QueryDSL, context: object, sort: SortParam}>>,
```

robinbisping:
As per discussion with Marc, we have added some type hints to the storage format but decided to not add any instructions on how to resolve li-document-search on your own to nudge customers to use the Composition API.

PR #824, opened 2024-04-12
🐞Fix search indexing

peyerluk:
# Changelog

* 🐞 Clean up text before it is indexed by lunr. This fixes a bug where any text directly adjacent to html tags would not be indexed (e.g. links, inline code blocks, most code).


DaRaFF (commented):
👍

PR #825, opened 2024-04-15
Custom table cell fixes

looer:
I recently had to implement a custom table cell for Handelsblatt and saw a couple of things in the docs which were not correct.


JSchenk8 (approved):
Nice one, thank you!

PR #834, opened 2024-04-29
Update Documentation for Document Teasers

christian-schmitz-ld:
# Motivation

NZZ wanted to test editable teasers, read the docs and reached out for help. I tried to setup an example configuration for them. Out of the given documentation I found it very difficult to understand the teaser guide. After digging deep and lots of testing I do now understand the implementation and therefore I want to fill gaps on the documentation as well as use meaningful identifiers (`teaser`, `teaser-include`, `teaser-service`...).

# Changelog

- 📖 Update documentation regarding Document Teasers
- 📖 Update `opener` useage for WSL 2 (Windows) in README.md


content/guides/documents/includes/document-teasers.md
```diff
@@ -83,13 +43,223 @@ Here is an example includes configuration to consider for the editable teaser us
...
+      async render (params, context) {
+        // params.article.reference.id contains the id of the linked document
+        // params.article.value contains part of the DocumentVersion ({systemdata, metadata}) 
+        // of the include (because preload: true is configured)
+        const documentVersion = params.article.value
+        return {
+          editableContent: !params.readOnly,
+          content: [{
+            id: `teaser-${documentVersion.systemdata.documentId}`,
+            component: 'my-teaser',
```
benib, line 153: With some newer projects, we started naming this component `teaser-template` to distinguish it from the `article-teaser` aka. `teaser-include`.

content/guides/documents/includes/document-teasers.md
```diff
@@ -83,13 +43,223 @@ Here is an example includes configuration to consider for the editable teaser us
...
+            content: {
+              image: parseImageData(documentVersion.metadata.teaserImage),
+              title: documentVersion.metadata.title,
+              lead: 'lead from include',
+              showByline: 'byline from include',
```
benib, line 158: In real world, those would come from the `documentVersion.metadata` as well, do you want to adjust this?

content/guides/documents/includes/document-teasers.md
```diff
@@ -83,13 +43,223 @@ Here is an example includes configuration to consider for the editable teaser us
...
+              link: 'https://example.com'
```
benib, line 159: I suggest to omit this one. It's confusing if the `teaser-template` contains a `doc-link` directive. If it has one, a user sees a UI to change this link, but she shouldn't. Adding a link to the teaser is normally done in the delivery rendering.

content/guides/documents/includes/document-teasers.md
```diff
@@ -83,13 +43,223 @@ Here is an example includes configuration to consider for the editable teaser us
...
+
+For a detailed description of the include api see [Includes Server API]({{< ref "/reference/document/includes/server-customization/" >}}).
+
+## Teaser Component
+We now have a registered include component `teaser-include` and service `teaser-service` and it is time to define our teaser component `my-teaser` which will be rendered within the document (replacing the include's placeholder markup). Rembember we already registerd `my-teaser` [here]({{< ref "#register-components" >}}) for our project.
+
+```js
+// my-teaser.js
+module.exports = {
+  name: 'my-teaser',
```
benib, line 222: name it `teaser-template` if you change it above.

content/guides/documents/includes/document-teasers.md
```diff
@@ -83,13 +43,223 @@ Here is an example includes configuration to consider for the editable teaser us
...
+    default: false
+  }, {
+    type: 'toggle',
+    name: 'showByline',
+    label: {en: 'show byline', de: 'Byline anzeigen'},
+    default: true
+  }],
+  properties: ['teaser-type'],
+  html: `
+    <a class="teaser" doc-link="link" style="display: block;">
```
benib, line 246: remove the `a` with the `doc-link` directive and change it to a `div` if you agree that it's a bit weird.

content/guides/documents/includes/document-teasers.md
```diff
@@ -83,13 +43,223 @@ Here is an example includes configuration to consider for the editable teaser us
...
+    type: 'toggle',
+    name: 'showLead',
+    label: {en: 'show lead', de: 'Lead anzeigen'},
+    default: false
+  }, {
+    type: 'toggle',
+    name: 'showByline',
+    label: {en: 'show byline', de: 'Byline anzeigen'},
+    default: true
+  }],
```
benib, line 243: Although this is nice as an example of how to use `toggle` directives, I have never seen this done like that in real world. Instead, different `teaser-template`-components would be used. So I suggest to omit this and make the example simpler with that.

content/guides/documents/includes/document-teasers.md
```diff
@@ -10,62 +10,22 @@ To drag and drop Document Cards onto a document to create Teaser Components you
...
+// teaser-include.js
+module.exports = {
   name: 'teaser-include',
   label: 'Teaser',
-  iconUrl: 'URL to an SVG icon',
+  iconUrl: 'path/to/an/icon',
   directives: [
     {
-      name: 'teaser',
+      name: 'teaser-include-dir-id',
```
benib, line 25: I'm slightly confused by the name `*-dir-id`. Is this an abbreviation for something?

benib (commented):
I have left some comments. I know that the example wasn't so clear before and you already improved it. While at it, you might want to consider my input.

content/guides/documents/includes/document-teasers.md
christian-schmitz-ld: Sure, I think import is only that they are unique within the guide. So you mean for `my-teaser` you started using `teaser-template`?

benib: Yes, this would be in line with what I have seen in recent new project configurations like `F.A.Z.`. I'm not up to date with the definitions and best practices you have in your team though, but it definitely makes sense to reflect them in the documentation.

content/guides/documents/includes/document-teasers.md
christian-schmitz-ld: Sure I will, there are side functionalities which I still don't understand in detail yet (like the doc-include). I used the upstream implementation from the service project (there is the teaser in use) and took it from there thinking this makes sense. Maybe we should also change the implementation on the upstream demo projects to keep everything aligned.

content/guides/documents/includes/document-teasers.md
christian-schmitz-ld: This I explicitly changed because it was very confusing having one more `teaser` identifier. As  I tested it I saw there is no impact in changing this id. Since it is a directive I just used *-dir-id. If it has an impact on the implementation then I really would like to understand.

benib: This will influence the data-model for the document. This is how the `content` of the document will look like:
```
content: [{
  component: 'teaser-include',
  content: {
    'teaser-include-dir-id': { <-- the name of the directive
       params: {
          // the params defined by the paramsSchema of the service
       },
       service: 'teaser-service'
    }
  }
}[
```

There is some impact on the implementation as it is defining the data model of the resulting document.

benib (changes requested):

benib (approved):
👏 

I would suggest to remove the `readOnly` functionality from the example here. This is really just there in the upstream service example project to test some behavior without having to change the code.

Otherwise this looks very nice and clean, thank you 👍

PR #839, opened 2024-04-30
fix(composition api): clarity on fields vs includes

JSchenk8:
Whilst using the composition API today I noticed that if you add `fields: ['includes']` and `resolveIncludes: true` then the content does not resolve the includes, they are only resolved in the separate array. 

I added this as a simple sentence to make sure people don't do what I did and spend ages thinking my include service wasn't resolving anything!


ajwild:
This could also be a bug. I don't know the intention of the Composition API code well enough.

JSchenk8:
@ajwild I can open a ticket if we want to dig into it? 🤔

christian-schmitz-ld (approved):

marcbachmann:
> then the content does not resolve the includes, they are only resolved in the separate array

What was expected here? Could you send me some examples? At the moment I don't think this is a bug. Let's still merge this.

PR #846, opened 2024-05-07
Add warning that document creation flows are incompatible with author mappings

robinbisping:


content/guides/editor/document-creation-flow.md
```diff
@@ -87,6 +87,10 @@ projectConfig.editorSettings = {
 
 When you go to the dashboard in the editor, you will see a "Create Breaking News" button.
 
+{{< info >}}
+Document Creation Flows don't support author prefilling. If author prefilling is configured, Livingdocs will default back to creation wizards.
```
benib, line 91: It's a bit of mess, sorry. I believe this needs some clarification to not be misleading.

`Document Creation Flows` really don't support automatic author prefilling. If anything needs to be done regarding author prefilling, it needs to be done manually. There are creative ways to do that [implemented at NPG](https://github.com/livingdocsIO/livingdocs-npg/blob/main/server/app/create-functions/article.js#L8-L25) for example. It is not using the "feature" author-prefilling, but achieves a similar result.

But: Document Creation Flows can still be used, even if author prefilling is configured. The fallback is implemented for the **automatic** creation flow. This is somehow just concerning our codebase. It's what happens if no Document Creation Flow is configured (on a dashboard) at all. So eventually, users/configurers don't even know that Document Creation Flows are a thing.

The fallback is there because we moved the create UI code from angular to vue and the author prefilling part is the only thing that is not yet supported in the new Vue create UI code.

benib (changes requested):

content/guides/editor/document-creation-flow.md
robinbisping: Thanks a lot for the explanation, Beni! I see. It came up during release testing (with the service project), that's why I created this PR. Users with an author mapping are shown the create wizards, users without a mapping see the Document Creation Flow.

Since this is only concerning our code base, it's probably ok to just close this PR to not confuse people unnecessarily, what do you think? It didn't seem to have been a problem in the past.

Alternatively, we could only add the warning to the author prefilling page. There, we could instruct users to configure Document Creation Flows to have a consistent behaviour for all users.

robinbisping: Sounds good, let's close it then. Thanks for the context.

PR #882, opened 2024-07-11
Separate Comyan deprecation and feature

JordiVM:
Take deprecation text from Comyan to a feature section. It makes more clear what is actually being deprecated.

### DIFF
--- a/content/operations/releases/release-2024-07.md
+++ b/content/operations/releases/release-2024-07.md
@@ -77,7 +77,7 @@ No prior preparations are required after rolling out this release.
 ### Migrate the Postgres Database :fire:
 No upstream migrations were added this release.
 
-{{< feature-info "Exposed functionality" "server" >}}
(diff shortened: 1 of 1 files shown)


content/operations/releases/release-2024-07.md
```diff
@@ -331,6 +289,50 @@ liServer.registerInitializedHook(() => {
...
+      mappings: [
+        {field: 'ImgName', metadataPropertyName: 'title'},
+        {field: 'IPTCCaption', metadataPropertyName: 'caption'},
+        {field: 'IPTCByLine', metadataPropertyName: 'source'},
+        {fullObject: true, metadataPropertyName: 'comyan'},
+      ]
+    }
+  }
+]
+```
```
JordiVM, line 318: I am unsure if we should specify where to configure this section here. It is not clear to me where they have to configure it. It looks like part of the `projecConfig` but I am not so sure how it changed from the previous release.

PR #887, opened 2024-08-20
Add a page about desknet to kordiam migration

looer:


ajwild (approved):
This looks good. If we're confident the new domain is stable then we could default release-2024-09 to kordiam.app and update this document later.

PR #889, opened 2024-08-30
Inbox Documentation

dfreier:
- Removed all documentation for multilist editor
- Inbox now available on articles and pages, for content types and media types
- We communicate we introduced `inbox.contentTypes` with release-2024-09, although it has been there for longer but was only used by t-online


benib (approved):
👍

PR #902, opened 2024-10-02
fix(import images): updated

JSchenk8:


content/reference/public-api/imports/images.md
```diff
@@ -73,7 +73,9 @@ POST api/v1/import/images
 
 --description--
 
-The image import does both create and update images. The import remembers the `externalId` / `systemName` pair and if an import matches an existing pair, it will update (Hint: consider how to rebuild the externalId when you want to update images). The image import in Livingdocs is asynchronous. You post a batch of images that you want to import and get back an id with which you can query later to get your result.
+The image import creates images. The image import in Livingdocs is asynchronous. You post a batch of images that you want to import and get back an id with which you can query later to get your result.
```
JordiVM, line 76:
```suggestion
The image import creates images in the Media Library. The image import in Livingdocs is asynchronous. You post a batch of images that you want to import and get back a jobId that you can use to retrieve the import job state.
```

JordiVM (approved):

content/reference/public-api/imports/images.md
JordiVM: I think it is worth mentioning a job is created, so that the naming convention is consistent with other endpoints.

PR #907, opened 2024-10-23
Support for DefaultAzureCredential in Azure Blob Storage

JordiVM:
@DaRaFF and @JSchenk8 added to provide CSO POV. If we think other members of CSO should be included as well, we can add them.


content/customising/server-configuration/storage/_index.md
```diff
@@ -136,15 +136,18 @@ storage: {
 
 ### Azure Blob Storage
 
-The Azure Blob Storage provider just needs a Storage Account name, container name, and generate a SAS token with read, write, create and delete access to the Azure Storage Account.
-Please consult the [detailed instructions]({{< ref "/customising/server-configuration/storage/azure-blob-storage" >}}) to generate the Shared Access Signature (SAS) for `sasToken` config parameter.
+The Azure Blob Storage provider configuration just needs a Storage Account name, Container name. There are multiple options for authentication for Azure Blob Storage:
```
marcbachmann, line 139: Never simplify a configuration process with the word `just`. There can always be complications
 
```suggestion
The Azure Blob Storage provider configuration needs a Storage Account name and Container name. There are multiple options for authentication for Azure Blob Storage:
```

content/customising/server-configuration/storage/_index.md
```diff
@@ -136,15 +136,18 @@ storage: {
...
+- generate a SAS token with read, write, create and delete access to the Azure Storage Account. Please consult the [detailed instructions]({{< ref "/customising/server-configuration/storage/azure-blob-storage" >}}) to generate the Shared Access Signature (SAS) for `sasToken` config parameter.
+- use `DefaultAzureCredential` to automatically select the most appropriate authentication method based on the available environment. Do not forget to define the appropriate RBAC role for the identity used to access the storage account.
```
marcbachmann, line 141: Do not mention `DefaultAzureCredential` that prominent. 
The option in our product would be to leave out the SAS Token, in which case the default provider will be used.

I'd also mention that that way processes will rotate tokens automatically.

DaRaFF (approved):

PR #914, opened 2024-10-31
Add documentation for inbox permission and webhooks user info

arddor:
Relations:
- https://github.com/livingdocsIO/livingdocs-server/pull/7422
- https://github.com/livingdocsIO/livingdocs-server/pull/7410
- https://github.com/livingdocsIO/livingdocs-editor/pull/9119
- https://github.com/livingdocsIO/livingdocs-editor/pull/9131

New:
- webhook user info
- inbox permission


content/operations/releases/release-2024-11.md
```diff
@@ -392,10 +392,28 @@ For more information on how to configure it, please refer to the [`li-unique-id`
...
+  ]
+}
+```
+
 
 {{< feature-info "Access Control Management" "server" >}}
 ### `inboxWrite` access right :gift:
 
+We are introducing an new `inbox` permission. This enables users to send contentTypes to the inbox without having write access for them. Users then need `read` and `inbox` access on the contentType to be able to search inboxes (read) and send something into the inbox (inbox). Setting the `update` permission contains the inbox permission.
+
```
JordiVM, line 415: It might be worth to mention, that permissions are not granted automatically when introduced. I think we never mentioned that anywhere and customers are super confused about that. When a new content type or permission scope is introduced it is disabled for all groups. So they should make sure to enable it for the groups that need it after the deployment introduces the option.

content/operations/releases/release-2024-11.md
```diff
@@ -392,10 +392,28 @@ For more information on how to configure it, please refer to the [`li-unique-id`
 {{< feature-info "Webhooks" "server" >}}
 ### Webhooks: User Actor Info :gift:
 
+We are adding a config to send more information about the user in webhook payloads. External systems can leverage this to understand who triggered an action. If enabled the `name` and `email` for user actors are included in the payload.
```
JordiVM, line 395:
```suggestion
We are adding a config to send more information about users in webhook payloads. External systems can leverage this to understand who triggered an action. If enabled the `name` and `email` for user actors are included in the payload. We do not recommend enabling this configuration unless your external systems require it.
```

Not sure if we should add all that. I don't like it when customers enable all options just because more is better.

PR #925, opened 2024-11-26
fix: fix typo in db migration name

castuche:
November release notes instruct to run this migration
`node db/manual-migrations/011-move-metadata.js `
It is probably actually referring to this one `011-move-revision-metadata`.

I updated it, waiting for your confirmation that this assumption was correct :)


marcbachmann (approved):
Ouch, yes. Looks like the PR description wasn't updated after renaming the file.
Thanks ❤️

PR #940, opened 2024-12-19
Add peiq release info

arddor:
- add release notes for peiq drag & drop integration


marcbachmann (changes requested):
Please also extend that page:
https://docs.livingdocs.io/reference/project-config/settings/#integrations

marcbachmann (approved):

PR #955, opened 2025-01-22
Apply prettier

marcbachmann:
image

Only some files left unchanged because of readability in the docs


package.json
```diff
@@ -34,5 +34,9 @@
     "branches": [
       "main"
     ]
+  },
+  "devDependencies": {
+    "prettier": "^3.4.2",
```
eileenoo, line 39: With prettier, you should always install the exact version -> Further reading https://prettier.io/docs/en/install.html#summary

package.json
```diff
@@ -34,5 +34,9 @@
...
+    "prettier-plugin-go-template": "^0.0.15"
```
eileenoo, line 40: I would assume it is a good practise to do the same with these plugins 🙏

eileenoo (commented):
Looks good ☺️ I just have the two comments and one question: do you also want to hide the prettier commit? If yes you could already add the `.git-blame-ignore-revs` file 🙌

marcbachmann:
> I just have the two comments and one question: do you also want to hide the prettier commit? If yes you could already add the .git-blame-ignore-revs file 🙌

Yes, I'd just do it as there's no disadvantage. But I wanted to add it in the next pr. No need to update the same file twice.

PR #956, opened 2025-01-22
Add warning about missing or random component ids in print flow function output

ajwild:
Relations:

- Related PRs:
  - https://github.com/livingdocsIO/livingdocs-demo/pull/460
  - https://github.com/livingdocsIO/livingdocs-server/pull/7640


fuenkchen (approved):
Looks good. It may be beneficial to add another example for that case, but I leave it up-to-you if you see any value in it.

PR #965, opened 2025-02-26
Restructure public api

marcbachmann:
### Internal Changelog
- ♻️ Move all endpoint declarations into separate yaml files
- ♻️ Add versioning to the endpoints

Using prod build:
`hugo server --buildFuture --openBrowser --navigateToChanged --disableFastRender`

image
image

Using dev build:
`npm start`
or
`hugo server --buildDrafts --buildFuture --openBrowser --navigateToChanged --disableFastRender`
image


content/_content.gotmpl
```diff
@@ -0,0 +1,86 @@
...
+
+{{ $minimumVersion := partial "api-version-sequence" "2025-03" }}
+
+{{ $weight := 0 }}
+{{ range (sort $.Site.Data.releases "key" "desc") }}
+  {{ $version := slicestr .key 8 }}
+  {{ $sequence := partial "api-version-sequence" $version }}
+  {{ if lt $sequence $minimumVersion }}{{ continue }}{{ end }}
+
+  {{ $.AddPage (dict
```
robinbisping, line 45: Discussed to solve this through variables instead of pages so that no empty pages need to be generated.

robinbisping (approved):
Looks good! Maybe we could forward this to Sven so that he can apply some finishing design touches to it. In Firefox, the dropdown looks different than in Chrome (see video below).

Also we have some style wobble, but that's just a side effect of doing the version switch client-side:

https://github.com/user-attachments/assets/a118e3d9-ac4a-49e7-aa32-d8b40f59cf3c

data/endpoints/get_latest_publication_beta.yaml
```diff
@@ -1,12 +1,23 @@
 title: Get Latest Publication
 beta: true
 
+endpointId: get_latest_publication
+
+apiVersion:
+  eq: beta
+
 history:
```
robinbisping, line 9: I find it suboptimal that the history is versioned. I think it would be nice if newer changes also appear on old API versions. But there might be other opinions.

It's definitely a good starting point and we will see how it evolves as more API versions are introduced over time.

PR #967, opened 2025-02-26
Update to correct values and versions

dfreier:
Ralph reported that we have an incorrect version in the releases.json: https://livingdocs.slack.com/archives/C4TGTEC3F/p1740551353043389?thread_ts=1740470994.998009&cid=C4TGTEC3F

I've updated them, along with the properties about maintainance etc.

### DIFF
--- a/data/releases.json
+++ b/data/releases.json
@@ -15,13 +15,13 @@
     "key": "release-2025-03",
     "name": "March 2025",
     "ref": "/operations/releases/release-2025-03.md",
-    "upcoming": true,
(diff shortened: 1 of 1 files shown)


DaRaFF (approved):

JordiVM:
Something has already been backported to March release.

JordiVM:
We cannot set the `upcoming`, `current`, `maintained` yet as that should be done in 2 weeks with the release announcement. If we change current we might break npm push :)

PR #983, opened 2025-03-18
Fix import documentation

marcbachmann:
Relations:
- Server PR: https://github.com/livingdocsIO/livingdocs-server/pull/7857

### DIFF
--- a/data/endpoints/media-library-serve-image.yaml
+++ b/data/endpoints/media-library-serve-image.yaml
@@ -32,7 +32,7 @@ openapi:
         - public-api:read
   tags:
     - Media Library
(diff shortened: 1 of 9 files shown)


data/endpoints/media-library-serve-image.yaml
```diff
@@ -16,11 +16,85 @@ query: |
...
+      required: true
+      schema:
+        type: string
+  responses:
+    '200':
+      description: OK
+      content:
+        image/*:
+          schema:
+            type: string
```
marcbachmann, line 53: openapi for non-json responses is a mess 🙈

PR #992, opened 2025-05-05
Add separate openApi page with links to all OpenAPI JSON files

peyerluk:
# Changelog

- Add a separate Open API page for the public API docs that is easier to find
- Add a list of all version specific Open API JSON documents

### DIFF
--- a/themes/hugo-docs/layouts/shortcodes/openapi-list.html
+++ b/themes/hugo-docs/layouts/shortcodes/openapi-list.html
@@ -1,0 +1,10 @@
+{{- $versions := hugo.Store.Get "api-versions" -}}
+{{- range $versions -}}
+  <!-- Sideeffect: publish json file at path /openapi-version.json -->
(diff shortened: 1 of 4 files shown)


themes/hugo-docs/layouts/shortcodes/openapi-list.html
```diff
@@ -0,0 +1,12 @@
+{{- $versions := hugo.Store.Get "api-versions" -}}
+{{- range $versions -}}
+  <!-- Sideeffect: publish json file at path /openapi-version.json -->
+  {{ $resource := (resources.FromString (printf "openapi-%s.json" .version) (partialCached "openapi-template.gotmpl" (dict "Site" $.Site "version" .version) .version)) -}}
+  {{ $resource.Publish }}
+  {{- if not .upcoming -}}
```
marcbachmann, line 6: You should remove this conditions as upcoming versions are excluded on the production build automatically.
We often work against the upcoming release and try to keep the whole functionality consistent.

PR #994, opened 2025-05-09
Mention Kordiam Rebranding in Public API Changelog

marcbachmann:
### Changelog
- Add desknet > kordiam url changes to public api changelog & release notes

image

### DIFF
--- a/content/reference/public-api/changelog/2025-05/kordiam-rebranding-remove-desknet-endpoints.md
+++ b/content/reference/public-api/changelog/2025-05/kordiam-rebranding-remove-desknet-endpoints.md
@@ -1,0 +1,43 @@
+---
+title: 'Kordiam Rebranding: Remove Desknet endpoints in newer versions'
+type: changelog-entry
+weight: 1
+
+change:
+  date: 2025-05
+  type: deprecation
+---
+
+The public API endpoints of the version `2025-05` do not support `desknet` in the path anymore.
+
+That means the following api prefixes should get changed:  
+`/api/v1/desknet/` > `/api/2025-05/kordiam/`  
+`/api/v1/desknet-integration/` > `/api/2025-05/kordiam-integration/`
+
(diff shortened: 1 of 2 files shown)


content/reference/public-api/changelog/2025-05/kordiam-rebranding-remove-desknet-endpoints.md
```diff
@@ -0,0 +1,55 @@
...
+
+✅ `POST` `/api/2025-03/desknet/element`  
+✅ `PUT` `/api/2025-03/desknet/element`  
+✅ `PUT` `/api/2025-03/desknet/element/:documentId`  
+✅ `DELETE` `/api/2025-03/desknet/element/:documentId`
+
+✅ `POST` `/api/2025-03/kordiam/element`  
+✅ `PUT` `/api/2025-03/kordiam/element`  
+✅ `PUT` `/api/2025-03/kordiam/element/:documentId`  
+✅ `DELETE` `/api/2025-03/kordiam/element/:documentId`
```
robinbisping, line 27: I would find it clearer if we removed the URLs for past versions that remain unchanged. Otherwise, there are a lot of URLs listed here. I would focus only on what's changing in the latest API version, rather than reiterating what stays the same in older versions.

```suggestion
The Desknet endpoints are still supported in the api versions `v1` till `2025-03`.
```

robinbisping (approved):
Very clear, great use of Emojis! 👏 I added one small suggestion.

content/reference/public-api/changelog/2025-05/kordiam-rebranding-remove-desknet-endpoints.md
marcbachmann: I also thought it's a bit much. I've listed them as not a lot of consumers are familiar with the new versioning. We can also make sure the text is better instead of promoting the old endpoints with a checkmark.

PR #1012, opened 2025-06-26
fix(comments): confirm not enabled one ditable teasers

JSchenk8:
We firmly disabled comments on editable teasers [here](https://github.com/livingdocsIO/livingdocs-editor/pull/9082)

It wasn't documented so I thought I'd add it quickly.


DaRaFF (approved):
👍

PR #1014, opened 2025-06-30
Add script to generate unified doc for AI

DaRaFF:
# Motivation

The generated code in MSP downstreams is much better if you provide our livingdocs documentation to the AI Agent. Everything can be done via a file MCP server that makes the process much more efficient.

# Changelog

Provide a script `bin/generate-unified-docs.js` which generates `unified-documentation.md` to be used as input for AI Agents.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 node_modules
 public
 resources
+unified-documentation.md
--- a/bin/generate-unified-docs.js
+++ b/bin/generate-unified-docs.js
@@ -1,0 +1,175 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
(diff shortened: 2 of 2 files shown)


mbrc (approved):
So many line comments, it almost feels like reading old Drupal code 😄 Not that that's necessarily a bad thing – talking about the line comments.

Didn't try the script, but the code looks good. One input: There's only one top-level try/catch, meaning that the whole process will fail if there's an issue with a single file. It might be better to simply skip such files and at least provide a partial result that might be useful immediately, instead of sending the dev on a debug journey to get any output.

marcbachmann:
Why don't you generate a json? AI can also better process that

PR #1114, opened 2026-02-25
Improve Notifications reference and guide and add Microsoft Teams channel

ajwild:
I looked into the Notifications after a [Slack request](https://livingdocs.slack.com/archives/C4TGTEC3F/p1771913700999779) for further information. I found a few inconsistencies so I have tried to tidy things up. I do not know enough about setting up a Microsoft Teams bot to add further details to this section.


ajwild:
After digging into this a bit I found that Teams is not supported and the code in the server is only a partial implementation.
https://livingdocs.slack.com/archives/C4TGTEC3F/p1777563871757209

PR #1132, opened 2026-03-30
Add usage log entry operations to Public API

ajwild:
- Added `addUsageLogEntry`, `updateUsageLogEntry` and `removeUsageLogEntry` operations to media library patch endpoint of the Public API
- Updated the technical release notes for release-2026-05
- Added an entry into the Public API changelog
- Fixed some invalid OpenAPI schema definitions so that the file will open again!


robinbisping (commented):
There is an additional file that probably should be updated: `/reference/public-api/changelog/`

PR #1142, opened 2026-04-30
Add Claude Skill: Write feature section for Technical Release Notes

eileenoo:
A skill to make writing the technical release notes easier at the end of the release. 
This does not create docs for the feature, but just adds a feature section to the Technical Release Notes.

Feedback is welcome! Try it out :) 

Prerequirements: Github & Notion MCP are setup 

## TODO

- [x] Add better examples
- [x] Add arguments to decide wether you want one link at a time or all at once
- [x] Move skill into skills folder

### DIFF
--- a/.claude/commands/write-trn-feature-section.md
+++ b/.claude/commands/write-trn-feature-section.md
@@ -1,0 +1,119 @@
+Write a Feature section entry for a Livingdocs Technical Release Note (TRN) and insert it into the correct release file. Use when a developer wants to document a new feature in `content/operations/releases/`.
+
+## Step 1: Gather inputs
+
+Ask **one question at a time** — wait for each answer before asking the next:
+
+1. **Release** — identifier in `release-YYYY-MM` format. Accept plain year/month and normalize it yourself.
+2. **GitHub PR URLs** — ask for the first URL. After receiving it, ask "Any more PRs? (say 'done' to continue)" and keep collecting until done.
+3. **Notion requirement URL** — original requirement page (if available)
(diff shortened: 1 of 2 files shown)


ajwild (approved):
I read through it last week and I think it makes sense. All of my features already had technical release notes written, so I haven't tested it out. I'm happy enough to merge it as it is, and I'll try it out over the next two months. We can make improvements and extensions when we notice output we don't like, or delete it again if we find that Claude can work its magic without the skill.

eileenoo:
> I read through it last week and I think it makes sense. All of my features already had technical release notes written, so I haven't tested it out. I'm happy enough to merge it as it is, and I'll try it out over the next two months. We can make improvements and extensions when we notice output we don't like, or delete it again if we find that Claude can work its magic without the skill.

Thanks for reading through it @ajwild . 
Basically there are two goals I wanted to reach with this skill:
#1  Make the process or writing the Release Notes faster, without having to "think" about every step on your own. 
#2 Have a more unified setup of the Release Notes. One of the outcomes of the TRN workshops, @dfreier and me worked on, was that the Feature Sections (as well as the rest) should follow the same setup. Short intro, link to the documentation, not to technical, flag if something is available automatically, etc. And using this skill to write them should be the first step of reaching this. 

If we find out that it doesn't make sense to do that with a skill, we can also have a template or something else for the feature sections. Let's find out!

arddor:
I tried it. It worked fine but we might reduce the initial steps. I think we could in one step pass multiple links and it could figure out notion is documentation and github PRs are relevant code. Alternatively I think you can also pass arguments to the skill.

eileenoo:
> I tried it. It worked fine but we might reduce the initial steps. I think we could in one step pass multiple links and it could figure out notion is documentation and github PRs are relevant code. Alternatively I think you can also pass arguments to the skill.

Thanks for the feedback :) In the beginning I had it with multiple links in one step and thoughts it's nicer with one at a time. But I like adding arguments, so that the user can choose! 
I will add it 🙏

PR #1144, opened 2026-05-04
fix: paginationFindConfig and li-language update

clement-knox:
1. Flags the need to configure the limit of the paginationFindConfig value to ensure all data records are displayed in dashboards.

Screenshot 2026-05-05 at 16 03 52

Will appear on: https://docs.livingdocs.io/customising/server-configuration/#documents

2. Makes explicit that li-language plugins have can a visible UI component in the editor

Screenshot 2026-05-05 at 16 16 45

### DIFF
--- a/content/customising/advanced/editor-configuration/display-filter.md
+++ b/content/customising/advanced/editor-configuration/display-filter.md
@@ -96,7 +96,10 @@ You can use Metadata Filters, wherever Display Filters are allowed. The example 
 }
 ```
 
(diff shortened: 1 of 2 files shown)


ajwild:
Thanks for taking a look at this. I would prefer a link to the `paginationFindConfig`, or some extra text, each time we mention "Shows a max of 1000 filter options".

ajwild:
The screenshots look good, but the code doesn't match.

clement-knox:
> The screenshots look good, but the code doesn't match.

Yup always helps if I actually commit the changes 😞 done now!

ajwild (approved):

PR #1156, opened 2026-05-29
AI Parsable Technical Release Notes

ajwild:


mbrc:
Hi @ajwild, I tested the release notes from this PR by running the release update skill on the DFV downstream. I ran the update skill once each with the old release notes and this AI-parsable version. The end result came out identical but the run itself went smoother with the new release notes.

**What worked well:**
1. The detect/fix sections gave the agent an exact config path to grep for and a concrete fix to apply, so it didn't have to infer everything from the prose.
2. Pulling `executeDocumentCommands` out of the features list into a proper deprecation meant it actually got flagged as a follow-up, whereas in the current notes it's kind of buried and got missed entirely.
3. The `li-target-length` mapping table made that fix deterministic, and as an aside it quietly fixes a small error in the current notes, which say `mode` where it should be `modes`.
4. The PR the skill generates links each finding back to the relevant release note entry. The :fire: emojis appended to the title of each breaking change make those anchor links more complicated to build than they should be. Dropping them in the new notes let the agent generate clean slugs without a round of guessing.

**A couple of suggestions:**
- The dot-path notation in the detect sections (e.g. `mediaTypes[].editor.dashboard`) tripped the agent up. It grepped for `.dashboard` and missed the `dashboard:` key, so it might be worth giving an actual grep pattern.
- It'd help to add the runtime code (e.g. `LIDEP080`) to each entry that emits one, so a warning that shows up during the test run can be matched straight back to the right release and its fix.

ajwild:
Thank you for the feedback @mbrc! I'm going to split this PR into smaller PRs so it's easier to review and I'll incorporate the two suggestions. I'm starting with the deprecations and breaking changes skills as I feel like they're the most import changes.

ajwild:
I'll close this now as a lot has been covered by other PRs. The remaining changes are updating .claude/skills/write-trn-feature-section/SKILL.md and improving the UI to make it easier to differentiate between sections.

PR #1162, opened 2026-06-17
Rewrite release-2026-05 breaking changes & deprecations in AI-parsable format

ajwild:
## What & why

Applies the AI-parsable TRN format to the **Breaking Changes** and **Deprecations** sections of `release-2026-05`, so an AI agent (or operator) can parse them into a concrete upgrade checklist.

> **Stacked on #1161** — base branch is `trn-skills-breaking-deprecations` (the skills + format reference). Review/merge that first; this PR's diff will shrink to just the content once it lands. Part of splitting up #1156.

## Changes

- 6 breaking changes + 4 deprecations rewritten into the `Detect` / `Fix` contract, with icon-free entry headings.
- **Greppable `Detect` lines** — a literal token to search for (not an abstract dot-path), plus before/after tables for the path-rename entries.
- **`**Code:**` lines** linking each entry to its `LivingdocsBreakingChange` / `LivingdocsDeprecation` log code, taken verbatim from the server source. Entries with no runtime code (direct `lib/` imports, usage-log `paramsSchema`) correctly omit it.
- `## Deprecations` heading gains `:warning:`.
- `publicApi.executeDocumentCommands()` moved from Features → Deprecations (it's an opt-in default-behaviour change, not a feature).
- Redis: removal corrected to `release-2026-09` (first announced `release-2026-03`), per `LIDEP079`.
- Reserved dashboard handle casing fixed to `imageCollections` to match the server.

## Notes

- The commit uses `fix()` so semantic-release triggers a patch/deployment on merge.
- The Features section is otherwise untouched — the full feature reformat is a separate, later change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1165, opened 2026-06-18
Render full TOC for release notes

ajwild:
Part of splitting up https://github.com/livingdocsIO/documentation/pull/1156.

Before:

image

After:

image

The [skills](https://github.com/livingdocsIO/documentation/pull/1162) will remove the emojis from titles so it'll look a bit cleaner in the future. You could test with https://github.com/livingdocsIO/documentation/pull/1162.

### DIFF
--- a/themes/hugo-docs/assets/elements/table-of-contents.scss
+++ b/themes/hugo-docs/assets/elements/table-of-contents.scss
@@ -39,5 +39,9 @@
       @extend .section-label;
       @extend .section-label--table-of-contents;
     }
+
+    ul ul ul a {
+      @extend %font-xs;
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


dfreier:
Let's try it

dfreier (approved):

ajwild:
If making the sidebar sections collapsible is important then I'll take another look in a week or two. At the moment it's quicker to scan what's in the release than it used to be, but it might be slower if you wanted to look for a specific feature (one more click but less to read/scroll).

eileenoo (commented):
I'm not sure about this one. As you said it adds more detail but at the same time it makes it harder to find/jump to a section.
I would do this together with collapsable under menus, in one take.

PR #1166, opened 2026-06-18
Move release notes newsletter form to bottom of overview page

ajwild:


dfreier:
Not sure if we should do this.

Pro: Dev's get easier to the list of releases without scrolling (depending of screen size)
Con: People might not discover the signup form

I lean towards keeping it at the top, but replace it with a link to https://livingdocs.io/en/newsletter

PR #1172, opened 2026-07-07
Keep inbox items after dropping them into a document

robinbisping:
- https://github.com/livingdocsIO/livingdocs-server/pull/9521
- https://github.com/livingdocsIO/livingdocs-editor/pull/11214

### DIFF
--- a/content/operations/releases/release-2026-07.md
+++ b/content/operations/releases/release-2026-07.md
@@ -280,6 +280,14 @@ Which options are available depends on the dashboard's asset type and your confi
 
 For more information, see the [Card Tags and Display Settings]({{< ref "/guides/media-library/media-library-setup/#card-tags-and-display-settings" >}}) documentation.
 
+### Keep Inbox Items After Dropping Them Into a Document
+
+By default, an inbox item is removed from the inbox as soon as it is dragged into a document. Some workflows call for the opposite. An editor researching images for an article might want to keep them in the inbox while trying out different options during production. For those setups the item should remain in the inbox after it has been placed.
+
+A new project setting `settings.inbox.keepItemsOnDrop` controls this behavior. When set to `true`, inbox items stay in the inbox after being dropped into a document.
+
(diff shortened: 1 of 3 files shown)


content/operations/releases/release-2026-07.md
```diff
@@ -226,6 +223,23 @@ Sometimes clearing image caches after a revoke or modification is handled by ext
 }
 ```
 
+### Keep Inbox Items After Dropping Them Into a Document
+
+By default, an inbox item is removed from the inbox as soon as it is dragged into a document. Some workflows call for the opposite. An editor researching images for an article might want to keep them in the inbox while trying out different options during production. For those setups the item should remain in the inbox after it has been placed.
+
+A new project setting `settings.inbox.keepItemsOnDrop` controls this behavior. When set to `true`, inbox items stay in the inbox after being dropped into a document.
+
+```js
```
eileenoo: I think we can remove the config here. It links to the documentation, where all the infos can be found on how to activate it :)

content/reference/project-config/content-types.md
```diff
@@ -898,6 +899,28 @@ using the `useDashboardColumns` property.
 }
 ```
 
+### Keep Items on Drop
+
+{{< added-in "release-2026-07" block >}}
+
+By default, items are removed from the inbox once they are dragged into the document content. Set the `inbox.keepItemsOnDrop` [project setting]({{< ref "/reference/project-config/settings" >}}) to `true` to keep items in the inbox after they are dropped, for example so an item can be placed multiple times.
```
eileenoo: Maybe write it rather like that: 

```suggestion
By default, items are removed from the inbox once they are dragged into the document content. Set the `inbox.keepItemsOnDrop` [project setting]({{< ref "/reference/project-config/settings" >}}) to `true` to keep items in the inbox after they are dropped, for example to place an item multiple times.
```

eileenoo (approved):

PR #1191, opened 2026-07-13
Only show h2 and h3 of TRN sections in TOC

ajwild:
Relations:
- Related PR: https://github.com/livingdocsIO/documentation/pull/1165

# Motivation

Before #1165 there was not enough detail, after #1165 there was too much. There is no need to show the Detect and Fix headings for example, so rendering the whole TOC was a step too far.

# Changelog

Before #1165:

image

After #1165:

image

Now:

image

I've also dropped the h1 headings because they are not used. They caused an unnecessary `<ul>` element to wrap the TOC.


eileenoo:
Niceee, I also put that on my list 😄

PR #1194, opened 2026-07-14
Improve Claude Documentation Skills

eileenoo:
# Motivation 

We continously want to improve our Claude skills for the documentation.

- Say "no" instead of "none" to skip input (as it's shorter)
- Repurpose + rename "write-trn-feature-section" into "document-feature" SKILL. This skill now also writes and adds documentation for a new feature and then links to it in the TRN feature section
- Add the option to create a branch, commit the changes and open the PR for the "document-feature" SKILL
- Review and update system requirements (if needed)


.claude/skills/document-feature/SKILL.md
```diff
@@ -0,0 +1,220 @@
...
+      handle: 'distributionDates',
+      type: 'li-distribution-dates',
+      ui: {
+        config: {
+          precision: 'date' // Options: 'datetime' (default) or 'date'
+        }
+      }
+    }
+
+For more information, see the [Distribution Dates]({{< ref "/reference/distribution-dates" >}}) documentation.
```
arddor: I think that reference should be:
```
reference/document/metadata/plugins/li-distribution-dates.md
```

arddor (approved):

.claude/skills/document-feature/SKILL.md
```diff
@@ -0,0 +1,222 @@
...
+
+If yes:
+
+1. **Branch.** If currently on `main`, create a branch (working changes carry over):
+   ```bash
+   git checkout -b docs/<short-feature-slug>
+   ```
+   If already on a feature branch, commit there.
+2. **Stage only the files this skill touched** — the documentation page(s) from Step 3 and the release file from Step 6. Check `git status` first; don't blanket-add unrelated changes.
+3. **Commit.** Use a `feat` prefix so a docs deployment is triggered (a prefix like `docs`/`chore` does not deploy):
```
ajwild, line 197: I mostly use a `fix` for content changes and reserve `feat` for changes to the overall website structure. In the end it doesn't matter though, but I'd be interested to hear what others are doing.

eileenoo: I always wondered, why mostly "fix" was used for commits in the documentation. I'm fine with both, though. Would be nice if we use something consistent. I'll add it to my list :)

PR #1203, opened 2026-08-05
Replace the image cropping tool with Pintura

robinbisping:
Relations:

- Editor PR: https://github.com/livingdocsIO/livingdocs-editor/pull/11359


content/guides/media-library/media-library-setup/index.md
```diff
@@ -232,22 +232,25 @@ projectConfig.designSettings: {
       handle: 'mobile',
       label: 'Mobile',
       description: 'this image will be shown on small screens',
-      // on small screens, having exact same ratios across all images is not that important
-      // we let the user pick one but change freely
+      // On small screens, having the exact same ratios across all images is not that
```
robinbisping, line 235: I didn't add `added-in` and `removed-in` labels for this feature. I think the references are scattered across too many places, and the changes are relatively minor, so the lack of clarity caused by documenting both the old and new behavior isn't worth it. I made the same decision for the editor config properties, because they are so non-essential that in my opinion it's not worth keeping them documented for another two releases. Also, I think we tend to forget to remove these outdated descriptions anyway.

If you would like to see them added anywhere, please let me know :)

ajwild (approved):

PR #1206, opened 2026-08-10
Cleanup release tags

ajwild:
- `removed-in` and `removedIn` should all be 2026-03 or later
- `deprecated-in` and `deprecatedIn` should all be 2026-03 or later, except:
  - Referenced designs
  - Declarative document copy
  - Import -> `rawDocument.livingdoc`
  - Import -> `publicationDate`
  - Server config -> `auth.accessTokenSecret`
- `added-in` and `addedIn` should all be 2026-03 or later


robinbisping (approved):
What a way to start the week 🤩 Thanks for cleaning it up!

PR #1208, opened 2026-08-12
Document the image processing limits

ajwild:
Documents the image processing limits added in these server PRs:

- https://github.com/livingdocsIO/livingdocs-server/pull/9851
- https://github.com/livingdocsIO/livingdocs-server/pull/9852
- https://github.com/livingdocsIO/livingdocs-server/pull/9853
- https://github.com/livingdocsIO/livingdocs-server/pull/9854
- https://github.com/livingdocsIO/livingdocs-server/pull/9855
- https://github.com/livingdocsIO/livingdocs-server/pull/9856
- https://github.com/livingdocsIO/livingdocs-server/pull/9865

Based on https://github.com/livingdocsIO/documentation/pull/1207, which introduces the section it extends — review that one first.

- Documents the new `mediaLibrary.images.processing.timeout` and `mediaLibrary.images.processing.maxConcurrentResizes` options.
- Adds a **Sizing the Processing Limits** section: how the two concurrency options relate to Node.js' thread pool, why `os.availableParallelism()` rather than `os.cpus().length` is the number to size against, and recommended values per container CPU limit.
- Describes `maxConcurrentProcesses` as covering an upload end to end, which is what it does as of server PR #9865.
- Notes that `uploadProcessingDirectory` now holds every transformed upload rather than only large animated ones, and corrects the volume sizing guidance: the worst case is 2GB rather than 300MB for the `maxFileSize: 100mb` that customers commonly configure.
- Adds a release-2026-09 feature entry covering all of the above plus the raised upload timeout, the shared job for concurrent requests for the same variant, and the one-thread-per-image change.


content/customising/server-configuration/_index.md
```diff
@@ -1217,6 +1243,53 @@ mediaLibrary: {
...
 
+#### Sizing the Processing Limits
+
+Image processing bounds its own memory. At startup it reads the memory limit of the container it runs in and reserves a share of it for the images it is working on, then admits each image against what that image needs — which is known from its metadata before a pixel of it is decoded. A small image runs alongside many others; a large one runs on its own. Nothing to configure, and nothing to keep in step with the container size.
+
+What is left to decide is how large an image you are willing to accept, and how much of the process' thread pool image work may take.
+
+##### What the memory goes on
+
+A still image is streamed rather than held in full, so its cost follows the size being produced rather than the size going in: an 18 megapixel photo becomes a 360px thumbnail in 8MB. Almost everything in a media library is a single frame.
```
JordiVM, line 1254: @ajwild this is confusing, the size it takes is the input one, not the output even when it is being streamed, 360px photo for 8MB sound like a lot.

ajwild: It was measured, but it also includes overhead from various operations. Most JPEGs stream the full width but only a few rows at a time, so the input held in memory is much smaller than the full image. I expect Sharp can pause the stream to avoid loading too much while processing, but I'll try to confirm it.

However, progressive JPEGs need to buffer the full input so the statement isn't accurate for them. It could add an additional 100MB in the example given. I'm not sure how specific to be with the numbers here. Maybe keeping things generic like "5x input pixels + 5x output pixels + 3MB overhead" is the best way to communicate it, with one large or many small tables with real-world examples.

I'll have a think about this. The whole PR needs reworking now that it's not part of the September release. We also need to decide if we drop any of the ~15 PRs before the text can be finalised. Thanks for bringing it up though, I'll definitely go through all the numbers and verify them once everything is in the main branch.

JordiVM: Yeah, generic numbers make it really confusing because all those words together would make me think of something else. The `"5x input pixels + 5x output pixels + 3MB overhead"` makes it clear

content/customising/server-configuration/_index.md
```diff
@@ -1217,6 +1243,53 @@ mediaLibrary: {
...
+
+The thread pool is the other ceiling, so the last column is only reached where `UV_THREADPOOL_SIZE` allows it.
+
+##### The thread pool
+
+Image processing runs on Node.js' thread pool, which is also what every file, DNS and compression operation in the process uses. Its default size is **4 threads**. An image takes a thread once one is free and then holds it for the whole operation — several seconds for a large animated image.
+
+Keep `UV_THREADPOOL_SIZE` a little above the container's CPU limit, which `os.availableParallelism()` reports. Do not use `os.cpus().length`: it reports the host's cores and ignores the container's CPU limit. The headroom stops file and DNS operations from queueing behind image processing, while a much larger pool mostly buys memory — for a screenful of uncached animated thumbnails, raising it from `4` to `16` took the time from 12.8s to 8.8s and the peak from 2.6GB to 8.2GB. The memory reservation is what holds that second figure down now, but the pool is still what decides whether the extra threads are worth having.
```
JordiVM, line 1283: Just tested it, it reports the same with both commands :) 
```
> os.availableParallelism()
2
> os.cpus().length
2
```
It shows the Container CPU limit

JordiVM:
In summary, we have to define a sensible `maxResolution` and `maxTotalResolution` that match the Memory limit of the container, and configure the storage mounted path on the container. What is the path used for that? I can't see a config in the docu.

ajwild:
It's `**`uploadProcessingDirectory`** (string, default os.tmpdir())` and `**`resizingDirectory`** (string, default os.tmpdir())`, which is most likely returning `/tmp` for a linux container. They can both point to the same mount point (as they do with `/tmp`) because we always work within a `nanoid` directory (e.g. `${localFolder}/${nanoid()}`).

content/customising/server-configuration/_index.md
ajwild: I expect that Kubernetes handles things better (some kind of system level isolation) than the testing I did with local docker containers.

